PR #122, opened 2019-07-18
Handle optional password in postgres datasource

zknill:
Update postgres connection string so that it does not include the
password parameter when the password is set to empty string.

This solves an issue where the 'password=' part of the connection string
would cause queries to be run against the default database 'postgres'
and not the database listed in config.


cube2222:
Hey, thanks for submitting the fix! Merging.

Issue #133, opened 2019-08-03
Add support for csv files without a header row


cube2222:
Just use names like col1,col2,col3...

cube2222:
This is fixed on master.

PR #136, opened 2019-08-07
WIP: CSV optional column names

JasiekChomiak:
The implementation seems to work, but I haven't tested it yet that much (and haven't written any tests) so it's WIP

This solves #133


README.md
```diff
@@ -122,7 +122,7 @@ JSON file in one of the following forms:
 CSV file seperated using commas. The first row should contain column names.
 ##### options:
 - path - path to file containing the data, required
-
+- columns - optional, defaults to true. If false then column names are col1, col2...
```
cube2222: Please change this to something like:
containsHeaderRow, or headerRow

storage/csv/datasource.go
```diff
@@ -58,9 +62,35 @@ func (ds *DataSource) Get(variables octosql.Variables) (execution.RecordStream,
...
+
+		columnCount := len(firstRow)
+
+		columns = make([]string, columnCount)
+
+		for i := 0; i < columnCount; i++ {
+			columns[i] = fmt.Sprintf("col%d", i+1)
+		}
+
+		_, err = file.Seek(0, 0) /* we must return to the beginning of file */
```
cube2222: Why? The first row was the header row, we don't want it anymore

JasiekChomiak: It's actually the opposite. There is no header row, so we read the first one, to determine how many columns there are (to create column names col1, col2...). Alternatively the first row could be passed to the stream as a []string, same as in Excel. Don't know which method is better tough

cube2222: As discussed in private, I think it's better to read and save the column list on first record read.

storage/csv/datasource.go
```diff
@@ -60,56 +64,107 @@ func (ds *DataSource) Get(variables octosql.Variables) (execution.RecordStream,
...
+	alias           string
+	aliasedFields   []octosql.VariableName
+	first           bool
+	hasColumnHeader bool
+}
+
+func (rs *RecordStream) Close() error {
+	err := rs.file.Close()
+	if err != nil {
+		return errors.Wrap(err, "Couldn't close underlying file")
```
cube2222: errors should start lowercase

PR #138, opened 2019-08-09
WIP: Fix casing bug in column names

JasiekChomiak:
Fixing the bug that wouldn't allow users to specify sources with
non-lowercase column names


cube2222:
#146 fixes this without the annoying conflicts.

PR #139, opened 2019-08-12
Record metadata

cube2222:


execution/record.go
```diff
@@ -75,9 +132,26 @@ func (r *Record) Equal(other *Record) bool {
...
+	}
+
 	return true
 }
 
+func (r *Record) IsUndo() bool {
+	return r.metadata.undo
+}
+
+func (r *Record) EventTimeField() octosql.Value {
```
woojtek: Is this used somewhere? Why not using it in line 73

execution/record.go
```diff
@@ -43,6 +90,16 @@ func (r *Record) Fields() []Field {
 			Name: fieldName,
 		})
 	}
+	if !r.metadata.eventTime.IsZero() {
+		fields = append(fields, Field{
+			Name: octosql.NewVariableName("sys.event_time"),
+		})
+	}
+	if r.metadata.undo {
```
woojtek: Maybe isUndo() here?

execution/record.go
cube2222: Fixed

execution/record.go
cube2222: Fixed

PR #160, opened 2020-01-10
WIP: Record protobuf

cube2222:


execution/distinct.go
```diff
@@ -116,6 +120,7 @@ func (rs *recordSet) Insert(r *Record) (bool, error) {
 	return false, nil
 }
 
+//TODO: should the hash include other fields as well?
 func HashRecord(rec *Record) (uint64, error) {
-	return hashstructure.Hash(rec.data, nil)
+	return hashstructure.Hash(octosql.GetValuesFromPointers(rec.Data), nil)
```
cube2222: Distinct shouldn't grab into internal fields.

execution/record.go
```diff
@@ -197,3 +189,7 @@ type RecordStream interface {
 var ErrEndOfStream = errors.New("end of stream")
 
 var ErrNotFound = errors.New("not found")
+
+func NewMetadata() *Metadata {
```
cube2222: Is this function necessary?

JasiekChomiak: yeah it isn't - removing it

execution/distinct.go
JasiekChomiak: changed it so that Hash is now a method implemented on record (record.Hash)

PR #161, opened 2020-01-11
like-regexp

JasiekChomiak:
Previous implementation of LIKE is now the relation REGEXP. 
Implemented the LIKE relation as is in SQL (without ESCAPE)

### DIFF
--- a/execution/relation.go
+++ b/execution/relation.go
@@ -3,6 +3,8 @@ package execution
 import (
 	"context"
 	"regexp"
+	"strings"
+	"unicode"
 
 	"github.com/pkg/errors"
 
@@ -148,6 +150,10 @@ func NewLike() Relation {
 	return &Like{}
 }
 
+const LikeEscape = '\\'
+const LikeAny = '_'
+const LikeAll = '?'
+
 func (rel *Like) Apply(ctx context.Context, variables octosql.Variables, left, right Expression) (bool, error) {
 	leftValue, err := left.ExpressionValue(ctx, variables)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


execution/relation.go
```diff
@@ -168,13 +174,74 @@ func (rel *Like) Apply(ctx context.Context, variables octosql.Variables, left, r
...
 	if err != nil {
 		return false, errors.Wrapf(err, "couldn't match string in like relation with pattern %v", rightValue)
 	}
 	return match, nil
 }
 
+//we assume that the escape character is '\'
+func likePatternToRegexp(pattern string) (string, error) {
+	length := len(pattern)
+	if pattern[length-1] == LikeEscape {
```
cube2222: Not true.

You can have a pattern like this: "\\"

execution/relation.go
```diff
@@ -168,13 +174,74 @@ func (rel *Like) Apply(ctx context.Context, variables octosql.Variables, left, r
...
+	sb.WriteRune('$') // match end
+
+	return sb.String(), nil
+}
+
+func isAlphaNumeric(r rune) bool {
+	return unicode.IsNumber(r) || unicode.IsLetter(r)
+}
+
+func needsEscaping(r rune) bool {
```
cube2222: dots need escaping

execution/relation.go
```diff
@@ -168,13 +174,74 @@ func (rel *Like) Apply(ctx context.Context, variables octosql.Variables, left, r
...
+	var sb strings.Builder
+
+	sb.WriteRune('^') // match start
+
+	for i, r := range pattern {
+		if r == LikeEscape { // if we find an escape sequence we just rewrite it
+			nextRune := pattern[i+1] // no illegal reference, because / can't be last
+
+			if nextRune == LikeAny || nextRune == LikeAll {
+				i += 1
```
cube2222: Use a normal loop if you modify i inside it. This doesn't work like you think it does. (i is by value in each loop iteration)

execution/relation.go
```diff
@@ -168,13 +174,74 @@ func (rel *Like) Apply(ctx context.Context, variables octosql.Variables, left, r
...
+		return "", errors.New("the pattern cannot end with an escape character")
+	}
+
+	var sb strings.Builder
+
+	sb.WriteRune('^') // match start
+
+	for i, r := range pattern {
+		if r == LikeEscape { // if we find an escape sequence we just rewrite it
+			nextRune := pattern[i+1] // no illegal reference, because / can't be last
```
cube2222: It's \ not /

execution/relation.go
```diff
@@ -168,13 +174,74 @@ func (rel *Like) Apply(ctx context.Context, variables octosql.Variables, left, r
...
+	length := len(pattern)
+	if pattern[length-1] == LikeEscape {
+		return "", errors.New("the pattern cannot end with an escape character")
+	}
+
+	var sb strings.Builder
+
+	sb.WriteRune('^') // match start
+
+	for i, r := range pattern {
```
cube2222: In my opinion you can just store a boolean "escaping", which gets set to true whenever you encounter a functioning escape character.

Now whenever you enter the loop and escaping=true, then just rewrite the current character literally.

If you exit the loop with escaping=true, error, because you have a functioning escape character as the last codepoint in the string.

execution/relation.go
JasiekChomiak: You are absolutely right, I will change that :)

execution/relation.go
```diff
@@ -148,6 +149,10 @@ func NewLike() Relation {
 	return &Like{}
 }
 
+const LikeEscape = '\\'
```
cube2222: those can be private

execution/relation.go
```diff
@@ -168,13 +173,78 @@ func (rel *Like) Apply(ctx context.Context, variables octosql.Variables, left, r
 			leftValue.Show(), rightValue.Show(), leftValue.GetType(), rightValue.GetType())
 	}
 
-	match, err := regexp.MatchString(rightValue.AsString(), leftValue.AsString())
+	patternString, err := likePatternToRegexp(rightValue.AsString())
+	if err != nil {
+		return false, errors.Wrapf(err, "couldn't transform LIKE pattern %v to regexp", patternString)
+	}
+
+	match, err := regexp.MatchString(patternString, leftValue.AsString())
```
cube2222: Maybe precompile regexp?

PR #162, opened 2020-01-12
Aggregates

woojtek:


streaming/aggregate/avg.go
```diff
@@ -0,0 +1,145 @@
...
+
+	var currentAvgNum octosql.Value
+	err = currentAvgNumStorage.Get(&currentAvgNum)
+	if err == storage.ErrKeyNotFound {
+		currentAvgNum = octosql.MakeInt(0)
+	} else if err != nil {
+		return errors.Wrap(err, "couldn't get current elements number for avg from storage")
+	}
+
+	newScalar := octosql.MakeFloat(float64(currentAvgNum.AsInt()) / float64(currentAvgNum.AsInt()-1))
```
cube2222: division by zero

streaming/aggregate/avg.go
```diff
@@ -0,0 +1,145 @@
...
+	case octosql.TypeFloat:
+		retractedComponent = octosql.MakeFloat(value.AsFloat() / float64(currentAvgNum.AsInt()-1))
+	default:
+		return errors.Errorf("unsupported value type passed to avg: %s", valueType)
+	}
+
+	currentAvgVal = octosql.MakeFloat(currentAvgValScaled.AsFloat() - retractedComponent.AsFloat())
+	currentAvgNum = octosql.MakeInt(currentAvgNum.AsInt() - 1)
+
+	if currentAvgNum.AsInt() > 0 {
```
cube2222: If new number is 0, you have to delete all used storage.

streaming/aggregate/avg.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+
+		err = currentAvgNumStorage.Set(&currentAvgNum)
+		if err != nil {
+			return errors.Wrap(err, "couldn't set current elements number for avg in storage")
+		}
+	} else {
+		err = currentAvgStorage.Clear()
+		if err != nil {
+			return errors.Wrap(err, "couldn't set current avg in storage")
```
cube2222: couldn't clear :)

streaming/aggregate/avg.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+
+		err = currentAvgNumStorage.Clear()
+		if err != nil {
+			return errors.Wrap(err, "couldn't set current elements number for avg in storage")
```
cube2222: same

streaming/aggregate/sum.go
```diff
@@ -28,7 +32,11 @@ func (agg *Sum) AddValue(ctx context.Context, tx storage.StateTransaction, value
 	case octosql.TypeFloat:
 		currentSum = octosql.MakeFloat(currentSum.AsFloat() + value.AsFloat())
 	case octosql.TypeDuration:
-		currentSum = octosql.MakeDuration(currentSum.AsDuration() + value.AsDuration())
+		if currentSum.GetType() == octosql.TypeInt { // No value set
```
cube2222: This looks kinda terrible. Also, what if the previous values _were_ ints?

cube2222: Different records can have different types in a field. We have to check for that and throw an error accordingly.

streaming/aggregate/max.go
```diff
@@ -0,0 +1,91 @@
...
+	return nil
+}
+
+func (agg *Max) GetValue(ctx context.Context, tx storage.StateTransaction) (octosql.Value, error) {
+	currentMaxStorage := storage.NewMap(tx.WithPrefix(currentMaxPrefix))
+
+	var currentMax octosql.Value
+	var currentMaxCount octosql.Value
+
+	it := currentMaxStorage.GetReverseIterator()
```
cube2222: Maybe this should instead be GetIterator(WithReverse())?

streaming/aggregate/sum.go
woojtek: fixed type checking for empty valueState case

woojtek: added type checking, now it returns error if current value type and sum value type doesn't match

streaming/aggregate/sum.go
```diff
@@ -41,49 +58,60 @@ func (agg *Sum) AddValue(ctx context.Context, tx storage.StateTransaction, value
...
 
+	valueType := value.GetType()
+
 	currentSum, err := agg.GetValue(ctx, tx)
 	if err != nil {
 		return errors.Wrap(err, "couldn't get current sum")
 	}
 
-	clearSum := true
+	// TODO - is this needed ?
```
cube2222: Yes, it's needed because if the retraction comes before the record there may be type error I think

streaming/aggregate/key.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (agg *Key) GetValue(ctx context.Context, tx storage.StateTransaction) (octosql.Value, error) {
+	currentKeyStorage := storage.NewValueState(tx.WithPrefix(currentKeyPrefix))
+	currentKeyCountsStorage := storage.NewValueState(tx.WithPrefix(currentKeyCountsPrefix))
+
+	var currentKeyCount octosql.Value
+	err := currentKeyCountsStorage.Get(&currentKeyCount)
+	if err == storage.ErrNotFound || currentKeyCount.AsInt() <= 0 {
+		return octosql.ZeroValue(), nil // what should be returned when there is no key in key storage?
```
cube2222: error or zero value, this shouldn't happen, I think?

streaming/aggregate/avg.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	err := agg.underlyingSum.AddValue(ctx, tx, value)
+	if err != nil {
+		return errors.Wrap(err, "couldn't add value to sum for avg")
+	}
+
+	err = agg.underlyingCount.AddValue(ctx, tx, value)
+	if err != nil {
+		err2 := agg.underlyingSum.RetractValue(ctx, tx, value) // TODO - is this necessary
```
cube2222: No, the transaction will be rolled back.

streaming/aggregate/avg.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	err := agg.underlyingSum.RetractValue(ctx, tx, value)
+	if err != nil {
+		return errors.Wrap(err, "couldn't retract value from sum for avg")
+	}
+
+	err = agg.underlyingCount.RetractValue(ctx, tx, value)
+	if err != nil {
+		err2 := agg.underlyingSum.AddValue(ctx, tx, value) // TODO - is this necessary
```
cube2222: no, as above

streaming/aggregate/avg.go
```diff
@@ -0,0 +1,109 @@
...
+		return octosql.ZeroValue(), errors.Wrap(err, "couldn't get current sum for avg")
+	}
+
+	currentCount, err := agg.underlyingCount.GetValue(ctx, tx)
+	if err != nil {
+		return octosql.ZeroValue(), errors.Wrap(err, "couldn't get current elements count for avg")
+	}
+
+	if currentCount.AsInt() <= 0 { // no values or early retractions
+		return octosql.MakeInt(0), nil // TODO - I don't think returning anything other than "neutral element" would be better
```
cube2222: Yes, I think this is ok.

cube2222: The layer above should make it so this doesn't happen anyways.

streaming/aggregate/aggregate_test.go
```diff
@@ -0,0 +1,72 @@
...
+	assert.NotNil(t, val)
+	assert.Equal(t, expected.GetType(), val.GetType())
+
+	switch expected.GetType() {
+	case octosql.TypeInt:
+		assert.Equal(t, expected.AsInt(), val.AsInt())
+	case octosql.TypeFloat:
+		assert.Greater(t, expected.AsFloat(), val.AsFloat()-octosql.MakeFloat(eps).AsFloat())
+		assert.Less(t, expected.AsFloat(), val.AsFloat()+octosql.MakeFloat(eps).AsFloat())
+		//assert.Equal(t, expected.AsFloat(), val.AsFloat())
```
JasiekChomiak: A tiny change: maybe you could just define eps as octosql.MakeFloat(epsilon_value).AsFloat() and then you wouldn't have to write it twice

streaming/aggregate/avg.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		return errors.Wrap(err, "couldn't add element to elements count for avg")
+	}
+
+	return nil
+}
+
+func isAppropriateType(valueType octosql.Type) bool {
+	return valueType == octosql.TypeInt || valueType == octosql.TypeFloat
```
JasiekChomiak: I think we can also calculate an average of Durations

streaming/aggregate/avg.go
JasiekChomiak: Yeah I think that's the best way to approach it

streaming/aggregate/avg.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if currentCount.AsInt() <= 0 { // no values or early retractions
+		return octosql.MakeInt(0), nil // TODO - I don't think returning anything other than "neutral element" would be better
+	}
+
+	var currentAvg octosql.Value
+
+	switch currentSum.GetType() {
+	case octosql.TypeInt:
+		currentAvg = octosql.MakeFloat(float64(currentSum.AsInt()) / float64(currentCount.AsInt()))
```
JasiekChomiak: Is there a specific reason you do float64(currentSum.AsInt()) instead of currentSum.AsFloat() ? (same goes for currentCount)

streaming/aggregate/avg.go
```diff
@@ -0,0 +1,109 @@
...
+	case octosql.TypeFloat:
+		currentAvg = octosql.MakeFloat(currentSum.AsFloat() / float64(currentCount.AsInt()))
```
JasiekChomiak: Same thing as above

streaming/aggregate/avg_test.go
```diff
@@ -0,0 +1,254 @@
...
+	AddValue(t, ctx, aggr, tx, octosql.MakeInt(2)) // Val: 14	Num: 4
+
+	ExpectValue(t, ctx, aggr, tx, octosql.MakeFloat(3.5))
+
+	// Wrong type passed
+	AddValueError(t, ctx, aggr, tx, octosql.MakeDuration(1234))
+
+	AddValueError(t, ctx, aggr, tx, octosql.MakeTime(time.Now()))
+
+	AddValueError(t, ctx, aggr, tx, octosql.MakeFloat(123.123)) // TODO - is this alright? You can't sum different types
```
JasiekChomiak: Yeah I think that's a good approach. If someone wants to sum mixed types they can just cast using the Float function

streaming/aggregate/avg_test.go
```diff
@@ -0,0 +1,254 @@
...
+	AddValue(t, ctx, aggr, tx, octosql.MakeFloat(0.9)) // Val: 10.2 Num: 4
+
+	ExpectValue(t, ctx, aggr, tx, octosql.MakeFloat(2.55))
+
+	// Wrong type passed
+	AddValueError(t, ctx, aggr, tx, octosql.MakeDuration(1234))
+
+	AddValueError(t, ctx, aggr, tx, octosql.MakeTime(time.Now()))
+
+	AddValueError(t, ctx, aggr, tx, octosql.MakeInt(123)) // TODO - is this alright? You can't sum different types
```
JasiekChomiak: yeah  - same as above

streaming/aggregate/first.go
```diff
@@ -3,23 +3,106 @@ package aggregate
...
+		err = currentFirstCountsStorage.Set(&value, &currentValueCount)
+		if err != nil {
+			return errors.Wrap(err, "couldn't set current value count in first storage")
+		}
+	}
+
+	return nil
 }
 
 func (agg *First) GetValue(ctx context.Context, tx storage.StateTransaction) (octosql.Value, error) {
```
JasiekChomiak: The approach with keeping count is a very nice idea, and it's intuitive once one understands what's happening (it becomes clear in GetValue). A comment in AddValue/RetractValue would clear things up ^^

streaming/aggregate/last.go
```diff
@@ -0,0 +1,110 @@
...
+var currentLastCountsPrefix = []byte("$current_last_counts$")
+
+type Last struct {
+}
+
+func NewLastAggregate() *Last {
+	return &Last{}
+}
+
+func (agg *Last) AddValue(ctx context.Context, tx storage.StateTransaction, value octosql.Value) error {
```
JasiekChomiak: Same comment applies as in first - if you put a comment here, reading the code the first time would be a bit easier :)

streaming/aggregate/max.go
```diff
@@ -0,0 +1,95 @@
...
+var currentMaxPrefix = []byte("$current_max$")
+
+type Max struct {
+}
+
+func NewMaxAggregate() *Max {
+	return &Max{}
+}
+
+func (agg *Max) AddValue(ctx context.Context, tx storage.StateTransaction, value octosql.Value) error {
```
JasiekChomiak: Wouldn't we want to only allow Max to store values of the same type? As in we wouldn't want someone to get Max from a float, int and a string, since it makes no sense. (I mean it's their fault that they have that sort of data, but maybe you could add some type checking here)

streaming/aggregate/min.go
```diff
@@ -0,0 +1,95 @@
...
+type Min struct {
+}
+
+func NewMinAggregate() *Min {
+	return &Min{}
+}
+
+var currentMinPrefix = []byte("$current_min$")
+
+func (agg *Min) AddValue(ctx context.Context, tx storage.StateTransaction, value octosql.Value) error {
```
JasiekChomiak: Same as above about the types

streaming/aggregate/sum.go
```diff
@@ -9,20 +9,37 @@ import (
...
 
+	valueType := value.GetType()
+
 	currentSum, err := agg.GetValue(ctx, tx)
 	if err != nil {
 		return errors.Wrap(err, "couldn't get current sum")
 	}
 
-	switch valueType := value.GetType(); valueType {
+	if isNeutralElement(currentSum) {
```
JasiekChomiak: This is kind of confusing... Does this mean we can sum floats, reach zero, and then start summing ints? That's what it looks like, since if we reach 0.0 for example, then we change currentSum to 0 (getAppropriateNeutralElement) and we can freely sum ints. I don't know if this was the desired effect

streaming/aggregate/aggregate_test.go
woojtek: sure, why not

streaming/aggregate/avg.go
woojtek: maybe, it was quite not obvious if we do, what do you think @cube2222 ?

streaming/aggregate/avg.go
woojtek: yea good point, that was needed before refactor, cause instead of Sum + Count there was Avg + Count (will fix this)
currentCount is always of TypeInt, so we have to typecast it

woojtek: some time ago i wrote that MakeInt(n).AsFloat() actually returns 0.0 for every positive n :P which I actually discovered in this exact place

streaming/aggregate/avg.go
woojtek: same as above :)

streaming/aggregate/max.go
woojtek: I can store the type of first incoming value and then check the type of it every time AddValue or RetractValue is called, but it might be one topic to discuss (@cube2222 )

streaming/aggregate/avg.go
cube2222: Yes, agreed.

streaming/aggregate/max.go
cube2222: Both approaches are valid really. If sorting between types sorts first by type (which it fits) then it kind of makes sense. But an error helps catch bugs, and I can't really think of a use case here.

streaming/aggregate/sum.go
woojtek: It's the thing I actually noticed (yet forgot to point out) and had pretty hard time to workaround. But it's a little bit different - You get appropriate neutral element only if GetValue returns MakeInt(0), which *should* happen only when there is no value in storage (ErrNotFound -> during first AddValue call). Which is a good idea, because types will match. BUT. Lets say the order looks like that:
1) Add(3)
2) Retract(3) (Here, the storage is NOT empty, but GetValue actually returns 0 which shouldn't be preceived as neutral element in this case, but as an actual sum of some elements)
3) Add(123.123) (This will work, because AddValue believes, that 0 returned by GetValue is a neutral element, and this function call is the first one -> which is actually not true)

This works only when Integers are summed in the first place, and I understand (and I did back when writing this) that it has to be fixed (e.x. by storing currentSum type in storage instead of asking "is current sum equal to neutral element". What do you think about it? The idea is quite the same as with "Min/Max type check" comment above
@cube2222 <--- another thing to discuss

streaming/aggregate/max.go
woojtek: sure, will add type checking then

streaming/aggregate/sum.go
cube2222: You can also have negative integers which summed up return zero.

streaming/aggregate/avg.go
woojtek: I'm sorry I misunderstood your comment. I need to typecast both currentSum and currentCount to floats in order to calculate currentAvg correctly. Therefore, this piece of code is truly needed.
@JasiekChomiak

streaming/aggregate/max.go
woojtek: The Master told us we don't have to bother :) @cube2222 @JasiekChomiak

streaming/aggregate/first.go
```diff
@@ -19,6 +19,9 @@ func NewFirstAggregate() *First {
 	return &First{}
 }
 
+// First storage contains Deque for actual order of elements added and Map for storing counts of every element.
+// Ones the element is added, we PushBack it into the Deque (because it is the last one to pick as 'first' in order).
```
JasiekChomiak: once*

streaming/aggregate/first.go
```diff
@@ -46,6 +49,8 @@ func (agg *First) AddValue(ctx context.Context, tx storage.StateTransaction, val
 	return nil
 }
 
+// Now, ones the element is retracted we don't know its current position in Deque, so the only thing we can do is
```
JasiekChomiak: once*

streaming/aggregate/last.go
```diff
@@ -19,6 +19,9 @@ func NewLastAggregate() *Last {
 	return &Last{}
 }
 
+// Last storage contains Deque for actual order of elements added and Map for storing counts of every element.
+// Ones the element is added, we PushFront it into the Deque (because it is the first one to pick as 'last' in order).
```
JasiekChomiak: once*

streaming/aggregate/last.go
```diff
@@ -46,6 +49,8 @@ func (agg *Last) AddValue(ctx context.Context, tx storage.StateTransaction, valu
 	return nil
 }
 
+// Now, ones the element is retracted we don't know its current position in Deque, so the only thing we can do is
```
JasiekChomiak: once*

PR #163, opened 2020-01-12
Add eventTimeField to physical node metadata.

cube2222:


physical/datasource.go
```diff
@@ -109,5 +109,5 @@ func (dsb *DataSourceBuilder) Materialize(ctx context.Context, matCtx *Materiali
 }
 
 func (dsb *DataSourceBuilder) Metadata() *metadata.NodeMetadata {
-	return metadata.NewNodeMetadata(dsb.Cardinality)
+	return metadata.NewNodeMetadata(dsb.Cardinality, "")
```
woojtek: different approach here than everywhere else (why is that so?) :
`return metadata.NewNodeMetadata(cardinality,  octosql.NewVariableName(""))`

physical/tvf.go
```diff
@@ -195,8 +195,17 @@ func (node *TableValuedFunction) Materialize(ctx context.Context, matCtx *Materi
...
+		return metadata.NewNodeMetadata(metadata.BoundedFitsInLocalStorage, octosql.NewVariableName(""))
+	case "tumble":
+		var cardinality metadata.Cardinality
+		source, err := node.getArgumentTable(octosql.NewVariableName("source"))
+		if err == nil {
+			cardinality = source.Metadata().Cardinality()
+		} else {
+			cardinality = metadata.BoundedFitsInLocalStorage
+		}
+		return metadata.NewNodeMetadata(cardinality, "window_end")
```
woojtek: raw string here? why not octosql.NewVariableName("window_end")

physical/datasource.go
cube2222: Fixed.

physical/tvf.go
cube2222: Fixed.

physical/map.go
```diff
@@ -51,6 +52,40 @@ func (node *Map) Materialize(ctx context.Context, matCtx *MaterializationContext
 	return execution.NewMap(matExprs, materialized, node.Keep), nil
 }
 
+func isVariableNameRecursive(expr Expression, name octosql.VariableName) bool {
```
JasiekChomiak: It’s not quite clear to me what this function does - a comment would be appreciated :)

physical/map.go
```diff
@@ -51,6 +52,40 @@ func (node *Map) Materialize(ctx context.Context, matCtx *MaterializationContext
...
+
 func (node *Map) Metadata() *metadata.NodeMetadata {
-	return metadata.NewNodeMetadata(node.Source.Metadata().Cardinality())
+	if node.Keep {
+		return metadata.NewNodeMetadata(node.Source.Metadata().Cardinality(), node.Source.Metadata().EventTimeField())
+	}
+
+	eventTimeField := node.Source.Metadata().EventTimeField()
+	var newEventTimeField octosql.VariableName
+	for i := range node.Expressions {
```
JasiekChomiak: Since you access node.Expressions[i] twice maybe for i, expr := would be just a bit more optimal

physical/map.go
cube2222: Fixed

physical/map.go
cube2222: Fixed.

PR #164, opened 2020-01-13
Proper int and float encoding

cube2222:


sorted_keys.go
```diff
@@ -27,9 +27,9 @@ const (
 )
 
 const (
-	NumberMarshalLength      = 1 + 1 + 8 // b[0] = type, b[1] = sign, b[2:] = marshal
-	BoolMarshalLength        = 1 + 1     // b[0] = type, b[1] = 0/1
-	NonexistentMarshalLength = 1         // null and phantom
+	NumberMarshalLength      = 1 + 8 // b[0] = type, b[1] = sign, b[2:] = marshal
```
cube2222: Fix comment

sorted_keys.go
```diff
@@ -192,61 +192,36 @@ func MonotonicUnmarshalPhantom(b []byte) error {
...
-	}
-	return binary.LittleEndian.Uint64(reverseByteSlice(b[2:])), nil
+func MonotonicUnmarshalInt64(b []byte) (int64, error) {
+	i := binary.BigEndian.Uint64(b[1:])
+	return int64(i) ^ math.MinInt64, nil
 }
 
 /* Marshal float */
-func MarshalFloat(f float64) []byte {
+func MonotonicMarshalFloat(f float64) []byte {
```
cube2222: Add comment that this is based on the rust implementation.

PR #167, opened 2020-01-16
deque

JasiekChomiak:
A modification of linked_list to support insertions and deletions from both sides
(so now it's a deque)


streaming/storage/deque.go
```diff
@@ -0,0 +1,383 @@
...
+
+	return ll.appendBytes(data, ll.firstElement, ll.decFirst)
+}
+
+//Adds new element to the end of the queue
+func (ll *Deque) PushBack(value proto.Message) error {
+	if !ll.initialized {
+		err := ll.initialize()
+		if err != nil {
+			return errors.Wrap(err, "failed to initialize in queue.PushBack()")
```
woojtek: PushBack() here while PushFront there -> different error message convention

streaming/storage/deque.go
```diff
@@ -0,0 +1,383 @@
...
+	return ll.pop(value, ll.lastElement-1, ll.decLast)
+}
+
+func (ll *Deque) pop(value proto.Message, index int, modifier func() error) error {
+	err := ll.peekIndex(value, index)
+	if err != nil {
+		return err
+	}
+
+	key := getIndexKey(index)
```
woojtek: The idea of deleting firstElement + 1 is confusing for me -> it means, that firstElement is pointing at first free index instead of first element (same with last element).

streaming/storage/deque.go
```diff
@@ -0,0 +1,383 @@
...
+
+	err = ll.tx.Delete(key)
+	if err != nil {
+		return errors.New("failed to delete element from queue in pop")
+	}
+
+	err = modifier()
+	return errors.Wrap(err, "failed to actualize queue attributes in pop")
```
woojtek: Difference to appendBytes function:
```
err = modifier()
if err != nil {
  return errors.Wrap(err, "failed to actualize queue attributes")
}

return nil
```

(Personally, don't really like wrapping possible nils :P . But let's just stick to 1 convention.)

streaming/storage/deque.go
```diff
@@ -0,0 +1,383 @@
...
+
+	key := getIndexKey(index)
+
+	data, err := ll.tx.Get(key)
+	if err != nil {
+		return errors.New("couldn't get value of element in peek")
+	}
+
+	err = proto.Unmarshal(data, value)
+	return errors.Wrap(err, "couldn't unmarshal the element in peek")
```
woojtek: same

streaming/storage/deque.go
woojtek: I think using
`return ll.pop(value, ll.lastElement, ll.decLast)`
instead of 
`return ll.pop(value, ll.lastElement-1, ll.decLast)`
would clarify this situation. This is the case in almost every function

streaming/storage/deque.go
```diff
@@ -0,0 +1,383 @@
...
+}
+
+func (ll *Deque) Print() error { //add initialize here (?)
+	var value octosql.Value
+	it := ll.tx.WithPrefix(dequeValueKeyPrefix).Iterator()
+
+	defer func() {
+		_ = it.Close() //TODO: panic here?
```
woojtek: why would you panic? :D

streaming/storage/deque.go
```diff
@@ -0,0 +1,383 @@
...
+	err := ll.decAttribute(dequeFirstElementKey)
+	if err != nil {
+		return errors.Wrap(err, "failed to decrease queue first attribute")
+	}
+
+	ll.firstElement -= 1
+	return nil
+}
+
+func (ll *Deque) incAttribute(attr []byte) error {
```
woojtek: I think you can move this function between incFirst and decLast so that it improve readability

streaming/storage/deque.go
JasiekChomiak: The problem is what happens when a queue is empty. At the beginning, an empty queue has it's first index at 0 and last index at 1, so it means the first "free" spots for PushFront and PushBack. Once you for example PushFront() then the first moves to -1, so to pop the first value, you actually have to add the +1. If we were to implement the first/last index as "position of first/last element" then it would become unclear what is to happen with an empty queue. I can try to reimplement it so it follows your suggestion and see how that goes

streaming/storage/deque.go
JasiekChomiak: This was a debug function I forgot to remove, sorry about that :3

streaming/storage/deque.go
woojtek: Well, if that's problematic then change struct field names e.x. firstElement -> firstFrontFreeSpot or smth :x

streaming/storage/deque.go
woojtek: Still you haven't corrected this specific place :P

streaming/storage/deque.go
```diff
@@ -0,0 +1,365 @@
...
+func (ll *Deque) peekIndex(value proto.Message, index int) error {
+	if ll.lastElement-ll.firstElement <= 1 {
+		return ErrEmptyQueue
+	}
+
+	key := getIndexKey(index)
+
+	data, err := ll.tx.Get(key)
+	if err != nil {
+		return errors.New("couldn't get value of element in peek")
```
cube2222: wrap?

streaming/storage/deque.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	return ll.peekIndex(value, ll.lastElement-1)
+}
+
+//we don't check if the queue has been initialized here, since we only
+//call peekIndex from PeekFront/PeekBack, which both initialize the queue
+func (ll *Deque) peekIndex(value proto.Message, index int) error {
+	if ll.lastElement-ll.firstElement <= 1 {
+		return ErrEmptyQueue
```
cube2222: Please use ErrNotFound

streaming/storage/deque.go
```diff
@@ -0,0 +1,392 @@
...
+	In a deque 1, 2, 3, 4 the Front is the 1 and the Back is the 4. This way
+	a Push/PopFront modify the front attributes, and Push/PopBack modify the back attributes.
+*/
+const (
+	defaultFirstFreeFrontSpot = 0
+	defaultFirstFreeBackSpot  = 1
+)
+
+var (
+	dequeValueKeyPrefix     = []byte("value")
```
cube2222: I think we should stick to one prefix naming convention. I've always been using $prefix$, here it's without the dollars, let's decide on one.

streaming/storage/deque.go
```diff
@@ -0,0 +1,392 @@
...
+
+	data, err := proto.Marshal(value)
+	if err != nil {
+		return errors.Wrap(err, "couldn't serialize given value")
+	}
+
+	return dq.appendBytes(data, dq.firstFreeFrontSpot, dq.decreaseFirstFreeFrontSpot)
+}
+
+//Adds new element to the back of the queue
```
cube2222: Your editor probably whines about this. Add a space after the beginning of a comment.

streaming/storage/deque.go
```diff
@@ -0,0 +1,392 @@
...
+//Removes the last element from the queue and returns it
+//Returns ErrNotFound if the list is empty
+func (dq *Deque) PopBack(value proto.Message) error {
+	//there is no need to call initialize here, since PeekBack calls initialize
+	err := dq.PeekBack(value)
+	if err != nil {
+		return err
+	}
+
+	return dq.pop(value, dq.firstFreeBackSpot-1, dq.decreaseFirstFreeBackSpot)
```
cube2222: Both pop and PeekBack use peekIndex under the covers. It's hard to understand what's happening here.

cube2222: I also think that those modifier shenanigans are kinda complex, and you could just call dq.decreaseFirstFreeBackSpot here, after using pop. It would probably work the same way, and would be much less complicated.

cube2222: It seems like you're calling it only so that initialize gets called?

streaming/storage/deque.go
```diff
@@ -0,0 +1,392 @@
...
+func (dq *Deque) Clear() error {
+	for dq.firstFreeBackSpot-dq.firstFreeFrontSpot > 1 {
+		key := getIndexKey(dq.firstFreeFrontSpot + 1)
+		err := dq.tx.Delete(key)
+
+		if err != nil {
+			return errors.Wrap(err, "couldn't clear the list")
+		}
+
+		//we call increaseFirstFree instead of simply dq.firstFree += 1 because
```
cube2222: If a machine dies during clear, then this structure will be created anew and the whole transaction won't have happened.

cube2222: If a machine dies during clear, then this structure will be created anew and the whole transaction won't have happened.

streaming/storage/deque.go
```diff
@@ -0,0 +1,392 @@
...
+	use the same one to add new values. In case they do we can set these to defaults.
+	*/
+	dq.firstFreeFrontSpot = defaultFirstFreeFrontSpot
+	dq.firstFreeBackSpot = defaultFirstFreeBackSpot
+	dq.initializedFront = false
+
+	return nil
+}
+
+func (dq *Deque) GetIterator() *DequeIterator {
```
cube2222: For a Deque it would make sense to provide a reverse iterator too.

streaming/storage/deque.go
```diff
@@ -0,0 +1,392 @@
...
+		return errors.Wrap(err, "failed to set new value")
+	}
+
+	return nil
+}
+
+func (dq *Deque) getAttribute(attr []byte) (int, error) {
+	value, err := dq.tx.Get(attr)
+	stringAttribute := string(attr)
+	switch err {
```
cube2222: Please just use a standard
if err == ErrNotFound {

} else if err != nil {

}
// nil case

streaming/storage/deque.go
JasiekChomiak: It actually doesn't, but I changed it :)

streaming/storage/deque.go
JasiekChomiak: True - I did just that

streaming/storage/deque.go
JasiekChomiak: I added the opts, so you can do it with WithReverse() now

streaming/storage/deque.go
```diff
@@ -0,0 +1,405 @@
...
+		if err != nil {
+			return errors.Wrap(err, "failed to initialize in PeekBack")
+		}
+	}
+
+	return dq.getValueAtIndex(value, dq.firstFreeBackSpot-1)
+}
+
+// Returns the value at given index.
+// We don't call initialize here, since this method is only called in methods like
```
cube2222: I think first and foremost, we don't call initialize here because we don't use any initializable values.

streaming/storage/deque.go
```diff
@@ -0,0 +1,405 @@
...
+	if err != nil {
+		return errors.Wrap(err, "couldn't unmarshal the element in peek")
+	}
+
+	return nil
+}
+
+// Clears all contents of the queue including the metadata.
+func (dq *Deque) Clear() error {
+	for dq.firstFreeBackSpot-dq.firstFreeFrontSpot > 1 {
```
cube2222: I think we need to call initalize here.

streaming/storage/deque.go
```diff
@@ -0,0 +1,405 @@
...
+	allOpts := []IteratorOption{WithDefault(), WithPrefix(dequeValueKeyPrefix)}
+	allOpts = append(allOpts, opts...)
+	it := dq.tx.Iterator(allOpts...)
+
+	return NewDequeIterator(it)
+}
+
+func (lli *DequeIterator) Next(value proto.Message) error {
+	err := lli.it.Next(value)
+	if err == ErrEndOfIterator || err == nil {
```
cube2222: I personally think it's better to use
```
if err == ErrEndOfIterator {
  ...
} else if err != nil {
   wrap
}

return nil
```

As everybody is used to the if handling the error case, and after that being the nil case, so it's less mental overhead.

streaming/storage/deque.go
JasiekChomiak: We use dq.firstFreeBackSpot

JasiekChomiak: We use dq.firstFreeBackSpot

streaming/storage/deque.go
JasiekChomiak: Yeah we actually do - will fix

streaming/storage/deque_test.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	if !areEqual {
+		log.Fatal("the iterator isn't correct")
+	}
+
+	_ = it.Close()
+
+	it = queue.GetIterator(WithReverse())
+	areEqual, err = TestDequeIterator(it, reverseValues(values))
```
woojtek: This is nice reverse iterator test, could You rewrite TestMapIterator in the same way?
I've done this in hurry and it's pretty dumb

JasiekChomiak: Will do

PR #168, opened 2020-01-19
Storage subscription

cube2222:


streaming/storage/storage.go
```diff
@@ -28,7 +27,7 @@ func NewBadgerStorage(db *badger.DB) *BadgerStorage {
 }
 
 func (bs *BadgerStorage) BeginTransaction() StateTransaction {
-	//bs.db.DropPrefix()
+	// bs.db.DropPrefix()
```
JasiekChomiak: If you are modifying the comment you might as well remove this line, I don't really think it's necessary to keep it there

streaming/storage/storage.go
```diff
@@ -45,15 +44,15 @@ func (bs *BadgerStorage) WithPrefix(prefix []byte) Storage {
...
-
-	err := bs.db.Subscribe(ctx, cb, bs.prefix)
-	if err != nil {
-		return errors.Wrap(err, "couldn't subscribe to db")
-	}
-
-	return nil
+func (bs *BadgerStorage) Subscribe(ctx context.Context) *Subscription {
+	return NewSubscription(ctx, func(ctx context.Context, changes chan<- struct{}) error {
+		return bs.db.Subscribe(ctx, func(kv *badger.KVList) error {
```
JasiekChomiak: Use KeyValueList defined above instead of *badger.KVList

streaming/storage/subscription.go
```diff
@@ -0,0 +1,68 @@
...
+	errs := make(chan error, 1) // Buffered so there won't be a goroutine leak if nobody reads the error.
+	sub := &Subscription{
+		cancel:  cancel,
+		changes: changes,
+		errors:  errs,
+		wg:      sync.WaitGroup{},
+	}
+
+	sub.wg.Add(1)
+	go func() {
```
JasiekChomiak: I think that a constructor that starts a goroutine is quite peculiar. Maybe it's a Go construct that I haven't seen. You could add a short comment explaining why you do it in this specific place.

streaming/storage/storage.go
cube2222: Autoformat, but sure

streaming/storage/storage.go
cube2222: removed the type instead

streaming/storage/subscription.go
cube2222: Added comment

PR #172, opened 2020-01-29
Thrift storage support

styczynski:
**Note! This PR does not contain the demo service as it requires a lot of automatically generated code (thrift generated client) that should not be reviewed. For clarity, I will create a separate PR with tests and the demo service.**

# Description

This PR adds support for Thrift storage.
Basically you can query a Thrift service if that service defined two methods:
```thrift
       i32 openRecords(),
       ReturnedRecord getRecord(1:i32 streamID),
```

In the future, it can be changed into a code that can make any arbitrary Thrift request.

# Basic functionality

The service that you are querying should return a new number when `openRecords()` is called. This number will identify the record stream. The service can return an arbitrary object when `getRecord()` is called with id obtained using `openRecords()`. When the service returns null then it notifies the client that the end of stream was reached.

Any arbitrary object can be returned as the record (all primitives all supported, lists are supported, but maps and sets are not yet).

If you want to get "real" field names instead of "table.field_0.field_1" you can set the path to the Thrift specification (see README.md inside storage/thrift directory).
The union fields, includes, and cyclic optional fields are not yet supported, but in basic cases, the resolver works out the names of the fields.

# Tests

There are basic tests written that instead of setting up a real HTTP server (as previously) just throw the data into buffers and feed the source with data.
The test set will be extended in the future.

# Plans for the future

In the future, we can add the ability to query any arbitrary Thrift method. It would be awesome if we could pass parameters to that call like `SELECT * FROM thrift.callSomeRemoteFunction("localhost:3000", 4, 0, "something")`


styczynski:
@cube2222  this contains pretty large amounts of code if you want I can split it somehow if I will have some time xD btw it would be awesome to work out some modular system for new storages as the codebase will grow - afaik go plugins are not supported anywhere except Linux distros so maybe it would be nice to work out other solution to have more modular plugin system for octosql?

cube2222:
> @cube2222 this contains pretty large amounts of code if you want I can split it somehow if I will have some time xD btw it would be awesome to work out some modular system for new storages as the codebase will grow - afaik go plugins are not supported anywhere except Linux distros so maybe it would be nice to work out other solution to have more modular plugin system for octosql?
> 
> @cube2222 You're the expert here, so I am jsut curious of your opinion

Thanks very much for the PR!

Nah, no need to split.

Yea, there is that problem with modules. I've been think about it too, but we can obviously just provide it for linux as a possibility. Other than that I think it's ok if static compilation is needed for now. We could also experiment with a Go interpreter like Yaegi to just parse the datasource code at runtime.

Anyways, I'll try to review this ASAP, but it may take a few days because I'm having exams now.

cube2222:
As said in real-live, the main thing this needs is a close function, for when we prematurely close the stream. (So there's no connection leak on either side).

PR #174, opened 2020-02-08
Add Triggers to SQL dialect

cube2222:


logical/group_by.go
```diff
@@ -110,5 +159,20 @@ func (node *GroupBy) Physical(ctx context.Context, physicalCreator *PhysicalPlan
 		}
 	}
 
-	return physical.NewGroupBy(source, key, node.fields, aggregates, node.as), variables, nil
+	triggers := make([]physical.Trigger, len(node.triggers))
+	for i := range node.triggers {
+		out, triggerVariables, err := node.triggers[i].Physical(ctx, physicalCreator)
+		if err != nil {
+			return nil, octosql.NoVariables(), errors.Wrapf(err, "couldn't get physical plan for trigger with index %d", i)
```
woojtek: I guess you could return nil instead of octosql.NoVariables() here as you do this everywhere else

logical/group_by.go
```diff
@@ -110,5 +159,20 @@ func (node *GroupBy) Physical(ctx context.Context, physicalCreator *PhysicalPlan
...
+		}
+
```
woojtek: there is no additional newline between Physical and MergeWith calls in lines 112 and 123, you can add a newline there :)

PR #175, opened 2020-02-12
Add suport for in-code defined SQL datasources

styczynski:
This PR adds the following features:
- Execute multiple SQL statements
- Create data sources using custom non-standard SQL statement (user don't need to create config file anymore)
- `octosql/` code and utilities to make Octosql more third-party friendly

**Important notes**:
- A huge part of changes in this MR is sql.go file so the real number of changed lines is invalid. There's no sensible way I could divide this MR in two.

## Detailed description

In my opinion, which may be biased, it's much quicker to copy just one (saved elsewhere) string containing SQL query than copying the configuration file and that string together.
This PR introduces custom `CREATE` statements:

```sql
  CREATE DATASOURCE [name] OF TYPE [type] OPTIONS [options_specs]
```

Where `[name]` is valid SQL identifier, `[type]` is an SQL literal string containing type name of the data source and `[options_specs]` is either empty or is of the form:
```
 option_specs := ( <option_name> <sql_expr> [, <identifier> <sql_expr>...] )
 option_name := <identifier> | <indentifier> . <option_name>
```

Example usage:
```sql
  CREATE DATASOURCE bikes OF TYPE 'json' OPTIONS (path 'bikes.json');
  SELECT * FROM bikes bikes
```

Configuration keys can be nested if needed:
```sql
  CREATE DATASOURCE some_source OF TYPE 'some_type' OPTIONS (x.y.z 'value');
```
Please note:
Nested values are supported, but arrays are not. The keys values are yet limited to SQL strings.

This is clean and expressive. The options notation was taken from [SQLServer OPTION CLAUSE](https://docs.microsoft.com/en-us/sql/t-sql/queries/option-clause-transact-sql?view=sql-server-ver15) and entire syntax was based on [CREATE SERVER](https://www.postgresql.org/docs/9.1/sql-createserver.html)

As the byproduct of that feature Octosql can now execute multiple queries in a single console command:

```bash
  $ octosql "SELECT * FROM bikes bikes; SELECT * FROM tests tests"
```

The other minor change was rewriting `app/app.go` into many files in `octosql/` which exports very simple API. This way the code is much cleaner and let's third party tools (like native bindings) use awesome Octosql features much easier.


octosql/run.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// Helper to run parsed AST node
+func (e *OctosqlExecutor) runParsedQuery(ctx context.Context, program *sqlparser.Program) error {
+	if program == nil {
+		return errors.New("Empty program cannot be executed.")
+	}
+
+	if program.Command.Statement != nil {
+		selectStmt, ok := program.Command.Statement.(sqlparser.SelectStatement)
```
cube2222: I think a type switch would be better here.

octosql/run.go
```diff
@@ -0,0 +1,143 @@
+package octosql
+
+import (
+	"context"
+	"fmt"
```
cube2222: Please use goimports with grouping to format this (this can be easily configured if you use goland).

parser/sqlparser/analyzer.go
```diff
@@ -221,72 +219,72 @@ type SetKey struct {
...
-		case *Default:
-			result[setKey] = "default"
-		default:
-			return nil, "", fmt.Errorf("invalid syntax: %s", String(expr))
-		}
-	}
-	return result, strings.ToLower(setStmt.Scope), nil
-}
+//func ExtractSetValues(sql string) (keyValues map[SetKey]interface{}, scope string, err error) {
+//	stmt, _, err := Parse(sql)
```
cube2222: I think you can just deleted this file if it's unused.

cube2222:
Overall I'm speechless, this is great! (though I still encourage you to speak with us (making an issue/request for comments before making PR's 😛)

Anyways, I've glanced through the code, put some cosmetic comments. I'll read it more thoroughly soon, but this is probably good to merge.

cmd/octosql/main.go
```diff
@@ -50,72 +30,19 @@ With OctoSQL you don't need O(n) client tools or a large data analysis system de
...
+			log.Fatal(err)
 		}
-		stateStorage := storage.NewBadgerStorage(db)
 
-		// Run query
-		err = app.RunPlan(ctx, stateStorage, plan)
+		err = executor.RunQuery(ctx, query)
 		if err != nil {
-			log.Fatal("couldn't run plan: ", err)
+			log.Fatal(err)
```
cube2222: I think providing context which part did the log.Fatal can be useful, so please bring that back. (the "couldn't run plan: ")

octosql/config.go
```diff
@@ -0,0 +1,39 @@
...
+		DataSources: []config.DataSourceConfig{},
+		Execution:   map[string]interface{}{},
+	}
+
+	if _, err := os.Stat(configPath); err == nil {
+		cfg, err = config.ReadConfig(configPath)
+		if err != nil {
+			return err
+		}
+	}
```
cube2222:
```
_, err := os.Stat(configPath)
if err == nil {
 	cfg, err = config.ReadConfig(configPath)
 	if err != nil {
 		return err
 	}
 } else if os.IsNotExist(err) {
} else {
    // error other than not exist handling
}
```

octosql/config.go
```diff
@@ -0,0 +1,39 @@
...
+
+	dataSourceRespository, err := physical.CreateDataSourceRepositoryFromConfig(
+		e.dataSourceFactories,
+		cfg,
+	)
+	if err != nil {
+		return err
```
cube2222: please use errors.Wrap and provide context

cube2222: This applies to all if err != nil { return err } 😉

octosql/handle_select.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/dgraph-io/badger/v2"
+	"log"
+)
+
+// This function executes all SQL select statements
+func (e *OctosqlExecutor) handleSelectStatement(ctx context.Context, statement sqlparser.SelectStatement) error {
+	// Parse input statement
+	plan, err := parser.ParseNode(statement)
+	if err != nil {
+		log.Fatal("couldn't parse query: ", err)
```
cube2222: It would be better to return errors from this function, it will fatal above anyways.

octosql/handle_select.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	opts := badger.DefaultOptions("")
+	opts.Dir = ""
+	opts.ValueDir = ""
+	opts.InMemory = true
+	db, err := badger.Open(opts)
+	if err != nil {
+		log.Fatal("couldn't open in-memory badger database: ", err)
```
cube2222: same here

parser/sqlparser/options_utils.go
```diff
@@ -0,0 +1,37 @@
+package sqlparser
+
+func injectEntry(v interface{}, entry OptionsSpecsEntry) interface{} {
```
cube2222: Please add a comment describing what this function does (I'm not sure I understand it)

parser/sqlparser/sql.y
```diff
@@ -613,6 +635,42 @@ create_statement:
     $$ = &DBDDL{Action: CreateStr, DBName: string($4)}
   }
 
+option_name:
+ table_id
```
cube2222: This looks like a hack, what was the problem with using just an identifier or something like that? (check out the table valued function argument names)

parser/sqlparser/sql.y
```diff
@@ -613,6 +635,42 @@ create_statement:
...
+  }
+| option_spec_entry_list ',' option_name '=' value_expression
+  {
+    $$ = append($$, OptionsSpecsEntry{
+        Key: $3,
+        Value: $5,
+    })
+  }
+
+options_specs:
```
cube2222: Please use the existing convention and change the name to options_specs_opt

cmd/octosql/main.go
styczynski: I will bring it back and do similar things for all the three `log.Fatal` statements.

parser/sqlparser/analyzer.go
styczynski: I forgot about this. Thanks!

octosql/run.go
styczynski: I changed the configuration. It should be fine now, but please let me know if the style still do not follow the project guidelines.

octosql/run.go
styczynski: Fixed.

octosql/config.go
styczynski: Oopsie, I forgot to change that. The error was discarded 😮 Thanks ❤️

octosql/config.go
styczynski: This will be fixed.

octosql/handle_select.go
styczynski: Fixed

octosql/handle_select.go
styczynski: Fixed

PR #176, opened 2020-02-14
Tests for streaming group by and many resulting fixes

cube2222:


execution/group_by.go
```diff
@@ -156,17 +187,20 @@ func (gb *GroupByStream) AddRecord(ctx context.Context, tx storage.StateTransact
 }
 
 var previouslyTriggeredValuePrefix = []byte("$previously_triggered_value$")
+var recordCountPrefix = []byte("$record_count$")
```
woojtek: declaration should be above AddRecord method

execution/group_by.go
```diff
@@ -176,20 +210,47 @@ func (gb *GroupByStream) Trigger(ctx context.Context, tx storage.StateTransactio
...
+			return nil, errors.Wrap(err, "couldn't clear previously triggered value")
 		}
 	}
-	output = append(output, NewRecordFromSlice(gb.outputFieldNames, values, opts...))
 
-	newPreviouslyTriggeredValuesTuple := octosql.MakeTuple(values)
-	err = previouslyTriggeredValues.Set(&newPreviouslyTriggeredValuesTuple)
-	if err != nil {
-		return nil, errors.Wrap(err, "couldn't set new previously triggered value")
+	if len(output) == 2 {
```
woojtek: You could add a comment here, not very clear for me what is going on :<

execution/record.go
```diff
@@ -29,6 +29,12 @@ func (id ID) Show() string {
 
 type RecordOption func(stream *Record)
 
+func WithClearUndo() RecordOption {
```
JasiekChomiak: I think maybe WithNoUndo() is a better name? Or maybe WithoutUndo()? The Clear is kinda confusing

execution/record.go
```diff
@@ -80,10 +86,11 @@ func NewRecordFromSlice(fields []octosql.VariableName, data []octosql.Value, opt
 }
 
 func NewRecordFromRecord(record *Record, opts ...RecordOption) *Record {
+	metadataCopy := *record.Metadata
 	r := &Record{
 		FieldNames: record.FieldNames,
 		Data:       record.Data,
-		Metadata:   &(*record.Metadata),
+		Metadata:   &metadataCopy,
```
JasiekChomiak: Out of curiosity - what is the reason behind dereferencing a pointer that is record.Metadata and then taking the new address? I mean I understand that we end up having the same value, but with a different address, but why?

execution/group_by.go
cube2222: added

execution/record.go
cube2222: Changed to WithNoUndo

execution/record.go
cube2222: Because this new metadata may be modified using options like WithNoUndo. This way we won't accidentally modify metadata from other record. We never modify data or fields, so we can reuse those without issue.

PR #177, opened 2020-02-14
Add stream IDs, proper storage prefixing, recovering from crashes.

cube2222:


execution/engine.go
```diff
@@ -133,37 +133,37 @@ func (engine *PullEngine) Run(ctx context.Context) {
 }
 
 func (engine *PullEngine) loop(ctx context.Context, tx storage.StateTransaction) error {
-	sourcePrefixedTx := tx.WithPrefix(engine.sourceStoragePrefix)
+	prefixedTx := tx.WithPrefix(engine.nodeStatePrefix)
 
-	watermark, err := engine.watermarkSource.GetWatermark(ctx, sourcePrefixedTx)
+	watermark, err := engine.watermarkSource.GetWatermark(ctx, tx)
```
JasiekChomiak: This seems like a very subtle change, yet there is nothing that explains it. Could you add a comment, as to explain why the watermark is suddenly created using the original transaction, and not the prefixed one, as was in the previous version?

execution/engine.go
```diff
@@ -133,37 +133,37 @@ func (engine *PullEngine) Run(ctx context.Context) {
...
 	if err != nil {
 		return errors.Wrap(err, "couldn't get current watermark from source")
 	}
 	if watermark.After(engine.lastCommittedWatermark) {
-		err := engine.irs.UpdateWatermark(ctx, tx, watermark)
+		err := engine.irs.UpdateWatermark(ctx, prefixedTx, watermark)
```
JasiekChomiak: Same thing here - these changes need commenting in my opinion

execution/engine.go
cube2222: I've added necessary comments.

execution/engine.go
cube2222: I've added necessary comments.

Qoziyev:
#329

PR #178, opened 2020-02-17
Modify limit to use state storage

cube2222:


execution/limit.go
```diff
@@ -68,19 +71,32 @@ func (node *LimitedStream) Close() error {
...
+	if limit.AsInt() > 0 {
 		record, err := node.rs.Next(ctx)
 		if err != nil {
 			if err == ErrEndOfStream {
-				node.limit = 0
+				limit = octosql.MakeInt(0)
 				return nil, ErrEndOfStream
 			}
 			return nil, errors.Wrap(err, "LimitedStream: couldn't get record")
 		}
```
JasiekChomiak: Newline here?

execution/limit.go
```diff
@@ -68,19 +71,32 @@ func (node *LimitedStream) Close() error {
...
-		node.limit--
-		if node.limit == 0 {
+		limit = octosql.MakeInt(limit.AsInt() - 1)
+		if limit.AsInt() == 0 {
 			node.rs.Close()
 		}
```
JasiekChomiak: Newline here?

execution/limit.go
```diff
@@ -68,19 +71,32 @@ func (node *LimitedStream) Close() error {
...
+		err = limitState.Set(&limit)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't set new limit state")
+		}
 		return record, nil
 	}
```
JasiekChomiak: I think it would improve our coherence of code if we agreed upon when is it appropriate to use multiple AsType functions, and when is it better to just define a value and use it everywhere. Here you use the limit.AsInt() function three times, so maybe it would be better to just define "limitValue := limit.AsInt()" and then use it everywhere? This is more about coding standards rather than optimizing the code, because this change is miniscule when it comes to performance

execution/limit.go
```diff
@@ -68,19 +71,32 @@ func (node *LimitedStream) Close() error {
...
+
+	if limit.AsInt() > 0 {
 		record, err := node.rs.Next(ctx)
 		if err != nil {
 			if err == ErrEndOfStream {
-				node.limit = 0
+				limit = octosql.MakeInt(0)
 				return nil, ErrEndOfStream
 			}
 			return nil, errors.Wrap(err, "LimitedStream: couldn't get record")
```
woojtek: wrap body starts with Uppercase

execution/limit.go
```diff
@@ -68,19 +71,32 @@ func (node *LimitedStream) Close() error {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't get limit state")
+	}
+
+	if limit.AsInt() > 0 {
 		record, err := node.rs.Next(ctx)
 		if err != nil {
 			if err == ErrEndOfStream {
-				node.limit = 0
+				limit = octosql.MakeInt(0)
```
woojtek: this variable is not used because it's followed by early return (?)

execution/limit.go
woojtek: same with wrap inside Close method

cube2222: That was like that previously, but yeah, I'll change it.

execution/limit.go
cube2222: It will be used in case someone tries to get a record again.

execution/limit.go
cube2222: Sure

execution/limit.go
cube2222: Sure

execution/limit.go
cube2222: This is definitely a case where a helper variable should be used. If it's used 3 times. I'll change it.

execution/limit.go
cube2222: Ah, I didn't save it.

PR #179, opened 2020-02-17
Modify union all to use state storage

cube2222:


execution/union_all.go
```diff
@@ -1,87 +1,120 @@
...
-	for {
-		firstRecord, err := node.first.Next(ctx)
-		if err != nil {
-			if err == ErrEndOfStream {
-				break
+	tx := storage.GetStateTransactionFromContext(ctx).WithPrefix(node.streamID.AsPrefix())
+	endOfStreamsMap := storage.NewMap(tx.WithPrefix(endsOfStreamsPrefix))
+
+	changeSubscriptions := make([]*storage.Subscription, len(node.sources))
+	for i := range node.sources { // TODO: Think about randomizing order.
```
JasiekChomiak: I think a more apt name instead of "i" would be something like "sourceIndex" or anything more descriptive

execution/union_all.go
```diff
@@ -1,87 +1,120 @@
...
+		key := octosql.MakeString(fmt.Sprint(i))
+		var endOfStream octosql.Value
+		err := endOfStreamsMap.Get(&key, &endOfStream)
+		if err == storage.ErrNotFound {
+		} else if err != nil {
+			return nil, errors.Wrapf(err, "couldn't get end of stream for source stream with index %d", i)
+		} else if err == nil {
+			// If found it means it's true which means there's nothing to read on this stream.
```
JasiekChomiak: I personally don't like the empty if statements, but the alternative is 
`if err != nil && err != storage.ErrNotFound {...} 
else if err == nil { continue }
`

execution/union_all.go
```diff
@@ -1,87 +1,120 @@
...
-				break
+	tx := storage.GetStateTransactionFromContext(ctx).WithPrefix(node.streamID.AsPrefix())
+	endOfStreamsMap := storage.NewMap(tx.WithPrefix(endsOfStreamsPrefix))
+
+	changeSubscriptions := make([]*storage.Subscription, len(node.sources))
+	for i := range node.sources { // TODO: Think about randomizing order.
+		key := octosql.MakeString(fmt.Sprint(i))
+		var endOfStream octosql.Value
+		err := endOfStreamsMap.Get(&key, &endOfStream)
+		if err == storage.ErrNotFound {
```
JasiekChomiak: Once again about the comments: before the whole `err := endOfStreamsMap...`
you could just add a comment "check if given stream source has already ended".
It would improve readability a lot

execution/union_all.go
```diff
@@ -1,87 +1,120 @@
...
+		} else if err != nil {
+			return nil, errors.Wrapf(err, "couldn't get end of stream for source stream with index %d", i)
+		} else if err == nil {
+			// If found it means it's true which means there's nothing to read on this stream.
+			continue
+		}
+
+		record, err := node.sources[i].Next(ctx)
+		if err == ErrEndOfStream {
```
JasiekChomiak: Here before the `record, err := ...` you could write "try to extract a record from the given source stream". I know you can extrapolate what happens in the code by reading it, but this short comment prepares the reader for what lies ahead and allows the first read-through to already target the actual logic instead of learning what the code does

execution/union_all.go
```diff
@@ -1,87 +1,120 @@
...
 		}
-		return firstRecord, nil
-	}
-	for {
-		secondRecord, err := node.second.Next(ctx)
-		if err != nil {
-			if err == ErrEndOfStream {
-				return nil, ErrEndOfStream
+
+		for j := 0; j < i; j++ {
```
JasiekChomiak: Here, a comment "managing subscriptions" (actually something more descriptive, I don't really get what's happening here :P)

streaming/storage/subscription.go
```diff
@@ -67,3 +69,49 @@ func (sub *Subscription) Close() error {
 		return err
 	}
 }
+
+func ConcatSubscriptions(ctx context.Context, subs ...*Subscription) *Subscription {
+	count := len(subs)
+
+	channels := make([]reflect.SelectCase, len(subs)*2+1)
```
JasiekChomiak: Ok this kind of length of slice magic deserves at least a small comment about what one is trying to achieve ;)

streaming/storage/subscription.go
```diff
@@ -67,3 +69,49 @@ func (sub *Subscription) Close() error {
 		return err
 	}
 }
+
```
JasiekChomiak: This function does some weird stuff, but I guess the code kinda has to look this way.

streaming/storage/subscription.go
woojtek: also, you have just calculated len(subs) 2 lines above so you can use 'count' variable here instead of calculating length once more

streaming/storage/subscription.go
```diff
@@ -67,3 +69,49 @@ func (sub *Subscription) Close() error {
...
+	for i := range subs {
+		channels[i] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(subs[i].changes)}
+	}
+	for i := range subs {
+		channels[count+i] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(subs[i].errors)}
+	}
+	return NewSubscription(ctx, func(ctx context.Context, changes chan<- struct{}) error {
+		channels[len(channels)-1] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(ctx.Done())}
+		var closedErr error
+	loop:
```
woojtek: wait what? isn't there any other way to handle this? :P

execution/union_all.go
cube2222: I don't really like the alternative :P We can discuss this further if you want.

execution/union_all.go
cube2222: I added something more descriptive.

streaming/storage/subscription.go
cube2222: No, and it's a standard pattern.
```
loop:
for {
  select {
  // ...
  break loop
  }
}
```
cause otherwise you'd be breaking out of the channel.

execution/union_all.go
JasiekChomiak: I don't like the alternative that much either, so we can stick with this one for the time being

execution/union_all.go
```diff
@@ -64,9 +63,12 @@ func (node *UnifiedStream) Next(ctx context.Context) (*Record, error) {
 
 	changeSubscriptions := make([]*storage.Subscription, len(node.sources))
 	for i := range node.sources { // TODO: Think about randomizing order.
```
JasiekChomiak: If you used sourceIndex above, then I think you should use it here as well

execution/union_all.go
```diff
@@ -64,9 +63,12 @@ func (node *UnifiedStream) Next(ctx context.Context) (*Record, error) {
...
-		key := octosql.MakeString(fmt.Sprint(i))
+		// Here we try to get a record from the i'th source stream.
```
JasiekChomiak: If you decide to change "i" to "sourceIndex" then don't forget to include the change in the comment please :D

streaming/storage/subscription.go
woojtek: alrighty

PR #181, opened 2020-02-19
Optimize triggers to fire many keys at once.

cube2222:
Optimize triggers to fire many keys at once.
Tidy up triggers in execution plan.
Remove dependency of trigger package on the execution package.

### DIFF
--- a/execution/trigger/counting.go
+++ b/execution/trigger/counting.go
@@ -7,7 +7,6 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/cube2222/octosql"
-	"github.com/cube2222/octosql/execution"
 	"github.com/cube2222/octosql/streaming/storage"
 )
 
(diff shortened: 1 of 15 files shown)


execution/trigger/multitrigger.go
```diff
@@ -43,39 +49,44 @@ func (m *MultiTrigger) UpdateWatermark(ctx context.Context, tx storage.StateTran
...
 					if err != nil {
-						return octosql.ZeroValue(), errors.Wrapf(err, "couldn't mark key fired in trigger with index %d after trigger with index %d fired", j, i)
+						return nil, errors.Wrapf(err, "couldn't mark keys fired in Trigger with index %d after Trigger with index %d fired", j, i)
 					}
 				}
 			}
-			return key, nil
-		} else if err != execution.ErrNoKeyToFire {
-			return octosql.ZeroValue(), errors.Wrapf(err, "couldn't poll key to fire in trigger with index %d: %v", i, m.triggers[i])
+			if outKeys == nil {
```
woojtek: I think you could just declare outKeys as
`
make([]octosql.Value, 0)
`

and then get rid of this if statement

execution/trigger/multitrigger.go
```diff
@@ -43,39 +49,44 @@ func (m *MultiTrigger) UpdateWatermark(ctx context.Context, tx storage.StateTran
...
+			} else {
+				outKeys = append(outKeys, keys...)
+			}
+		} else if err != nil {
+			return nil, errors.Wrapf(err, "couldn't poll keys to fire in Trigger with index %d: %v", i, m.triggers[i])
 		}
 	}
 
-	return octosql.ZeroValue(), execution.ErrNoKeyToFire
+	return outKeys, nil
```
woojtek: same, change from ErrNoKeyToFire to nil is not very clear for me

execution/trigger/watermark.go
```diff
@@ -104,18 +103,10 @@ func (wt *WatermarkTrigger) PollKeyToFire(ctx context.Context, tx storage.StateT
...
-		return octosql.ZeroValue(), execution.ErrNoKeyToFire
-	}
-
-	key, sendTime, err := timeKeys.GetFirst()
-	if err != nil {
-		if err == storage.ErrNotFound {
-			panic("unreachable")
-		}
-		return octosql.ZeroValue(), errors.Wrap(err, "couldn't get first key by time")
+		return nil, nil
```
woojtek: same

execution/trigger/watermark.go
```diff
@@ -124,35 +115,45 @@ func (wt *WatermarkTrigger) PollKeyToFire(ctx context.Context, tx storage.StateT
...
+			panic("unreachable")
+		}
+		return nil, errors.Wrap(err, "couldn't get keys by time until watermark")
 	}
 
-	err = timeKeys.Delete(key, sendTime)
-	if err != nil {
-		return octosql.ZeroValue(), errors.Wrap(err, "couldn't delete key")
+	for i := range keys {
+		if watermark.Before(times[i]) {
```
JasiekChomiak: If it's truly unreachable (as when one adds this as a default in a switch) then why add this if statement?
Is this like a sanity check?

execution/trigger/counting.go
```diff
@@ -65,51 +64,57 @@ func (ct *CountingTrigger) UpdateWatermark(ctx context.Context, tx storage.State
...
 
-func (ct *CountingTrigger) PollKeyToFire(ctx context.Context, tx storage.StateTransaction) (octosql.Value, error) {
+func (ct *CountingTrigger) PollKeysToFire(ctx context.Context, tx storage.StateTransaction) ([]octosql.Value, error) {
 	toSend := storage.NewValueState(tx.WithPrefix(toSendPrefix))
 	var out octosql.Value
 	err := toSend.Get(&out)
 	if err != nil {
 		if err == storage.ErrNotFound {
-			return octosql.ZeroValue(), execution.ErrNoKeyToFire
+			return nil, nil
```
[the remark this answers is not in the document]
cube2222: It's a new pattern in the whole package. Before you had octosql.Value being returned, which isn't a pointer so can't be returned as nil.

Now we are returning a slice, a nil slice is just an empty slice.

execution/trigger/counting.go
```diff
@@ -65,51 +64,57 @@ func (ct *CountingTrigger) UpdateWatermark(ctx context.Context, tx storage.State
...
+		return nil, errors.Wrap(err, "couldn't get value to send")
 	}
 
 	err = toSend.Clear()
 	if err != nil {
-		return octosql.ZeroValue(), errors.Wrap(err, "couldn't clear value to send")
+		return nil, errors.Wrap(err, "couldn't clear value to send")
 	}
-	return out, nil
+	return []octosql.Value{out}, nil
```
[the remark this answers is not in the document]
cube2222: That function doesn't work like that. AsSlice gets a slice from a tuple. out is not a tuple, it's a single value.

execution/trigger/multitrigger.go
cube2222: nil and this make are equal. This is just a small premature optimisation. I think without append this will be faster.

cube2222: No unnecessary copying.

execution/trigger/multitrigger.go
cube2222: As above.

execution/trigger/watermark.go
cube2222: Yeah, it's a sanity check.

cube2222: Cause I think this would be hell to debug if it didn't work.

execution/trigger/timesortedkeys.go
```diff
@@ -58,6 +58,50 @@ func (tsk *TimeSortedKeys) Update(key octosql.Value, t time.Time) error {
...
+		}
+
+		outTimes = append(outTimes, tuple[0].AsTime())
+		outValues = append(outValues, tuple[1])
+	}
+	if err == storage.ErrEndOfIterator {
+	} else if err != nil {
+		return nil, nil, errors.Wrap(err, "couldn't get first element from iterator")
+	}
+	if err := iter.Close(); err != nil {
```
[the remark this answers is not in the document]
cube2222: Sure.

PR #182, opened 2020-02-24
Watermark generator

woojtek:


execution/group_by.go
```diff
@@ -20,9 +21,10 @@ type Aggregate interface {
 }
 
 type GroupBy struct {
-	storage storage.Storage
-	source  Node
-	key     []Expression
+	storage   storage.Storage
+	source    Node
+	watermark Node
```
cube2222: GroupBy shouldn't have a watermark source.

The PullEngine above it does. Please look at how an execution group by is created from a physical groupBy.

The pull engine will pull watermarks from the watermark source, and update the watermark in the processbykey node. Which will trigger the necessary keys afterwards.

cube2222: No changes to execution/group_by.go should be necessary. Only to how the physical plan injects the watermark source into the pull engine.

execution/tvf/watermark.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, errors.Wrap(err, "couldn't get source record")
+	}
+
+	// TODO - ustaw nowa wartosc watermarka
+	timeValue := srcRecord.Value(s.timeField)
+	if timeValue.GetType() != octosql.TypeTime {
+		return nil, fmt.Errorf("couldn't get time field '%v' as time, got: %v", s.timeField.String(), srcRecord.Value(s.timeField))
+	}
+
+	currentWatermark, err := s.GetWatermark(ctx, tx) // TODO - nie mam tu transakcji - zrobic nowa (?)
```
cube2222: You can get a transaction out of the context, check out nodes like Limit.

cube2222: Ah, I've seen you've done that in the next file.

physical/tvf.go
```diff
@@ -187,6 +187,35 @@ func (node *TableValuedFunction) Materialize(ctx context.Context, matCtx *Materi
...
+		}
+
+		return tvf.NewZeroWatermark(matSource), nil
+
+	case "watermark":
+		source, err := node.getArgumentTable(octosql.NewVariableName("source"))
+		if err != nil {
+			return nil, err
+		}
+		timeField, err := node.getArgumentDescriptor(octosql.NewVariableName("time_field"))
```
cube2222: You can get the timefield from a record. There's a function for that.

woojtek: yea but that's the way to do this in tumble/range so i copied it

execution/tvf/watermark.go
woojtek: yea I realised that but didn't make a commit

physical/tvf.go
woojtek: what about this:

`timeField := source.Metadata().EventTimeField()`

and keep the tvf/watermark file unchanged?

cube2222: Tumble is more general. This could be an optional argument.

Yes, your proposal looks good.

execution/tvf/watermark.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/cube2222/octosql"
+	"github.com/cube2222/octosql/docs"
+	"github.com/cube2222/octosql/execution"
+	"github.com/cube2222/octosql/streaming/storage"
+)
+
+type Watermark struct {
```
cube2222: This is a WatermarkGenerator, not a Watermark

execution/tvf/watermark.go
```diff
@@ -0,0 +1,109 @@
...
+	ws := &WatermarkStream{
+		source:    sourceGetOutput.RecordStream,
+		timeField: w.timeField,
+	}
+
+	// This watermark stream is the new "latest" watermark source
+	return execution.NewGetOutput(ws, ws, nil)
+}
+
+type WatermarkStream struct {
```
cube2222: Same

physical/tvf.go
```diff
@@ -187,6 +187,32 @@ func (node *TableValuedFunction) Materialize(ctx context.Context, matCtx *Materi
 
 		return tvf.NewTumble(matSource, timeField, matWindowLength, matWindowOffset), nil
 
+	case "zero":
```
cube2222: yes indeed, this shouldn't be a tvf, but the default. (and be returned in the todo in the range tvf)

execution/order_by.go
```diff
@@ -51,24 +51,24 @@ func isSorteable(x octosql.Value) bool {
...
 	}
 
 	orderedStream, err := createOrderedStream(ctx, ob.expressions, ob.directions, variables, sourceStream)
 	if err != nil {
-		return nil, errors.Wrap(err, "couldn't create ordered stream from source stream")
+		return nil, nil, errors.Wrap(err, "couldn't create ordered stream from source stream")
 	}
 
-	return orderedStream, nil
+	return orderedStream, execOutput, nil
```
cube2222: Order By waits to read the whole stream. (it doesn't make much sense in streaming situations, as it doesn't return anything until it gets the whole stream and returns it as a whole). Maybe with an exception when we're grouping by event_time + something (this would need some restructure). But for now I think this should return a zero watermark.

execution/tvf/watermark_generator.go
```diff
@@ -0,0 +1,107 @@
...
+
+	ws := &WatermarkGeneratorStream{
+		source:    source,
+		timeField: w.timeField,
+	}
+
+	return ws, execution.NewExecOutput(ws), nil // watermark generator stream now indicates new watermark source
+}
+
+type WatermarkGeneratorStream struct {
```
cube2222: This watermark generator currently uses the maximum record seen as a watermark. This is a special case of the watermark generator we wanted to support which uses the maximum - some interval. For example, the watermark is the max record seen minus 10 seconds (so we expect lack of order no bigger than 10 seconds in our input stream)

cube2222: So please just modify this one to accept such a parameter.

physical/tvf.go
```diff
@@ -187,6 +187,27 @@ func (node *TableValuedFunction) Materialize(ctx context.Context, matCtx *Materi
 
 		return tvf.NewTumble(matSource, timeField, matWindowLength, matWindowOffset), nil
 
+	case "watermark":
```
cube2222: Maybe change the name to something like max_diff_watermark?

execution/tvf/watermark_generator.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (w *WatermarkGenerator) Get(ctx context.Context, variables octosql.Variables, streamID *execution.StreamID) (execution.RecordStream, *execution.ExecOutput, error) {
+	tx := storage.GetStateTransactionFromContext(ctx)
+	sourceStreamID, err := execution.GetSourceStreamID(tx.WithPrefix(streamID.AsPrefix()), octosql.MakePhantom())
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "couldn't get source stream ID")
+	}
+
+	source, _, err := w.source.Get(ctx, variables, sourceStreamID) // we don't need execOutput here since we become new watermark source
```
cube2222: become _the_ new (comment fix)

execution/tvf/watermark_generator.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "couldn't get source")
+	}
+
+	offset, err := w.offset.ExpressionValue(ctx, variables)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "couldn't get watermark offset")
+	}
+	if offset.GetType() != octosql.TypeDuration {
+		return nil, nil, errors.Errorf("invalid watermark offset: %v", offset)
```
cube2222: invalid watermark offset _type_. And after the value, please add the result of offset.GetType() too.

physical/tvf.go
```diff
@@ -221,6 +245,15 @@ func (node *TableValuedFunction) Metadata() *metadata.NodeMetadata {
...
 		return metadata.NewNodeMetadata(cardinality, octosql.NewVariableName("window_end"))
+	case "max_diff_watermark":
+		var cardinality metadata.Cardinality
+		source, err := node.getArgumentTable(octosql.NewVariableName("source"))
+		if err == nil {
+			cardinality = source.Metadata().Cardinality()
+		} else {
+			cardinality = metadata.BoundedFitsInLocalStorage
+		}
+		return metadata.NewNodeMetadata(cardinality, octosql.NewVariableName("time_field"))
```
cube2222: That's not true, we don't set a "time_field" variable. Just pass what you get from underneath.

execution/tvf/watermark_generator.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, errors.Wrap(err, "couldn't get source record")
+	}
+
+	timeValue := srcRecord.Value(s.timeField)
+	if timeValue.GetType() != octosql.TypeTime {
+		return nil, fmt.Errorf("couldn't get time field '%v' as time, got: %v", s.timeField.String(), srcRecord.Value(s.timeField))
+	}
+
+	// watermark value stored equals to (max_record_time - offset) that's why we multiply offset by -1
+	timeValueWithOffset := timeValue.AsTime().Add(s.offset * -1)
```
JasiekChomiak: I think visually `-1 * s.offset` looks better since you avoid the * - combo

execution/execution.go
```diff
@@ -2,15 +2,38 @@ package execution
...
+	"time"
 
 	"github.com/cube2222/octosql"
 	"github.com/cube2222/octosql/streaming/storage"
 
 	"github.com/pkg/errors"
 )
 
+// This struct represents additional metadata to be returned with Get() and used recursively (like WatermarkSource)
+type ExecOutput struct {
```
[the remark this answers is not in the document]
woojtek: sure

PR #186, opened 2020-02-26
Describe query as graphviz graph

cube2222:


physical/limit.go
```diff
@@ -45,3 +46,10 @@ func (node *Limit) Materialize(ctx context.Context, matCtx *MaterializationConte
 func (node *Limit) Metadata() *metadata.NodeMetadata {
 	return metadata.NewNodeMetadata(metadata.BoundedFitsInLocalStorage, node.Source.Metadata().EventTimeField())
 }
+
+func (node *Limit) Visualize() *graph.Node {
+	n := graph.NewNode("Limit")
+	n.AddChild("limit", node.LimitExpr.Visualize())
+	n.AddChild("source", node.Source.Visualize())
```
woojtek: I think that mostly "source" child is added in the first place, so maybe swap those 2 lines

physical/offset.go
```diff
@@ -45,3 +46,10 @@ func (node *Offset) Materialize(ctx context.Context, matCtx *MaterializationCont
 func (node *Offset) Metadata() *metadata.NodeMetadata {
 	return metadata.NewNodeMetadata(node.Source.Metadata().Cardinality(), node.Source.Metadata().EventTimeField())
 }
+
+func (node *Offset) Visualize() *graph.Node {
+	n := graph.NewNode("Offset")
+	n.AddChild("offset", node.OffsetExpr.Visualize())
+	n.AddChild("source", node.Source.Visualize())
```
woojtek: Same as with Limit

physical/requalifier.go
```diff
@@ -44,3 +45,10 @@ func (node *Requalifier) Metadata() *metadata.NodeMetadata {
 
 	return metadata.NewNodeMetadata(node.Source.Metadata().Cardinality(), eventTimeField)
 }
+
+func (node *Requalifier) Visualize() *graph.Node {
+	n := graph.NewNode("Requalifier")
+	n.AddField("qualifier", node.Qualifier)
+	n.AddChild("source", node.Source.Visualize())
```
woojtek: Same

physical/physical.go
```diff
@@ -175,6 +199,12 @@ func (le *LogicExpression) Materialize(ctx context.Context, matCtx *Materializat
 	return execution.NewLogicExpression(materialized), nil
 }
 
+func (le *LogicExpression) Visualize() *graph.Node {
+	n := graph.NewNode("Logic Expression")
+	n.AddChild("source", le.Formula.Visualize())
```
woojtek: in physical.Filter field of type Formula got "predicate" name, and here is "source"

physical/limit.go
cube2222: I'm trying to write them in an intuitive order which is based on how they're used.

If you have a limit, you first check it, and then look into the source.

If you have a map, you first get a record and then transform it.

physical/requalifier.go
cube2222: Here you have a field and child they'll be in different places in the node.

physical/physical.go
cube2222: Indeed, I tried to use the name "source" a lot, to not confuse users.

PR #188, opened 2020-03-01
Percentile Watermark Generator

woojtek:


execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,148 @@
...
+	if timeValue.GetType() != octosql.TypeTime {
+		return nil, fmt.Errorf("couldn't get time field '%v' as time, got: %v", s.timeField.String(), srcRecord.Value(s.timeField))
+	}
+
+	tx := storage.GetStateTransactionFromContext(ctx)
+	watermarkStorage := storage.NewValueState(tx.WithPrefix(percentileWatermarkPrefix))
+	eventsStorage := storage.NewDeque(tx.WithPrefix(percentileWatermarkEventsPrefix))
+
+	var watermarkPlace int       // represents position of event in sorted list that will become new watermark
+	if s.eventsSeen < s.events { // no need to pop first element from deque
```
cube2222: I think as long as we don't have at least s.events eventsSeen, we don't want to send any watermark. (Because we have too little data)

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,148 @@
...
+	err = eventsStorage.PushBack(&timeValue)
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't push back new event to events deque")
+	}
+
+	// TODO - well, do something clever to save new watermark value ...
+	// ...
+	// 1) easiest idea: store events in Map and then iterate through all <watermarkPlace> values to find watermark (slow)
+
+	newWatermark := octosql.MakeTime(time.Time{})
```
cube2222: Just a thing to keep in mind, watermarks can't decrease.

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,148 @@
...
+
+	tx := storage.GetStateTransactionFromContext(ctx)
+	watermarkStorage := storage.NewValueState(tx.WithPrefix(percentileWatermarkPrefix))
+	eventsStorage := storage.NewDeque(tx.WithPrefix(percentileWatermarkEventsPrefix))
+
+	var watermarkPlace int       // represents position of event in sorted list that will become new watermark
+	if s.eventsSeen < s.events { // no need to pop first element from deque
+		s.eventsSeen++
+
+		watermarkPlace = s.percentile * s.eventsSeen / 100
```
JasiekChomiak: I know it's the same thing math-wise, but I think `(s.percentile * s.eventsSeen) / 100` looks a bit better.

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,148 @@
...
+	} else {
+		watermarkPlace = s.percentile * s.events / 100
```
JasiekChomiak: Same as above

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,148 @@
...
+		timeField:  w.timeField,
+		events:     events.AsInt(),
+		percentile: percentile.AsInt(),
+		eventsSeen: 0,
+	}
+
+	return ws, execution.NewExecOutput(ws), nil // watermark generator stream now indicates new watermark source
+}
+
+type PercentileWatermarkGeneratorStream struct {
```
JasiekChomiak: Could you add a short comment that explains how this generator works? I understand that the `events` field is user given and is the amount of records that one suspects (?) and the `percentile` is after what percentile a watermark should be sent, but I am not sure I understand it 100%

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,148 @@
...
+		return nil, fmt.Errorf("couldn't get time field '%v' as time, got: %v", s.timeField.String(), srcRecord.Value(s.timeField))
+	}
+
+	tx := storage.GetStateTransactionFromContext(ctx)
+	watermarkStorage := storage.NewValueState(tx.WithPrefix(percentileWatermarkPrefix))
+	eventsStorage := storage.NewDeque(tx.WithPrefix(percentileWatermarkEventsPrefix))
+
+	var watermarkPlace int       // represents position of event in sorted list that will become new watermark
+	if s.eventsSeen < s.events { // no need to pop first element from deque
+		s.eventsSeen++
```
JasiekChomiak: Not sure that's the case here, but shouldn't `eventsSeen` be kept on disk in case of machine failure, so that the generator can restore itself correctly?

cube2222: You could just use the deque length I think?

woojtek: There is no method to extract deque length

cube2222: Add it then.

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	if events.GetType() != octosql.TypeInt || events.AsInt() < 1 {
+		return nil, nil, errors.Errorf("invalid watermark events: %v", events)
+	}
+
+	percentile, err := w.percentile.ExpressionValue(ctx, variables)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "couldn't get watermark percentile")
+	}
+	if percentile.GetType() != octosql.TypeInt || percentile.AsInt() < 1 || percentile.AsInt() > 99 {
```
JasiekChomiak: I think maybe it looks just a bit better when values such as this 1 and 99 are named. Maybe add constants 
`minPercentileValue` and `maxPercentileValue`? (I'm open for this discussion about this one tho)

JasiekChomiak: Or maybe this is an overkill - I dunno

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil, nil, errors.Errorf("invalid watermark percentile: %v", percentile)
+	}
+
+	frequency, err := w.frequency.ExpressionValue(ctx, variables)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "couldn't get watermark frequency")
+	}
+	if frequency.GetType() != octosql.TypeInt || frequency.AsInt() < 1 {
```
JasiekChomiak: If the above applies, then same here (maybe it is an overkill...) :P

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't add newest event to storage")
+	}
+
+	eventsLength, err := eventsStorage.Length()
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't get events deque length")
+	}
+
+	// In this situation we need to remove oldest event => this doesn't occur only when the oldest event seen is the first ever and length = 5 (update watermark but DON't remove oldest one)
```
JasiekChomiak: What is "5"? 
Also DON'T*

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,297 @@
...
+	if eventsLength > s.events { // TODO - which is more readable: this or 'eventsLength == s.events + 1' ?
```
JasiekChomiak: I think this one is more readable, since it suggests that if we have too many events stored, we get rid of the oldest one stored :)

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,297 @@
...
+				return nil, errors.Wrap(err, "couldn't get next value from events count map iterator")
+			}
+
+			eventsAlreadySeen += value.AsInt()
+
+			if eventsAlreadySeen >= watermarkPosition { // we've passed specified percentile of all events
+				break
+			}
+		}
+		err = eventsIterator.Close()
```
JasiekChomiak: Newline between '}' and `err = ...`?

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+
+		// Setting new watermark value
+		oldWatermark, err := s.GetWatermark(ctx, tx)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't get old watermark value from storage")
+		}
+
+		newWatermark := octosql.MakeTime(key.AsTime())
+		if newWatermark.AsTime().After(oldWatermark) { // watermarks can't decrease
```
JasiekChomiak: I am not sure about one thing - when we get the needed amount of events we recalculate the watermark's value. If the result we got is SMALLER than the current watermark value we don't update the watermark (so basically nothing happens watermark-value-wise), but yet we remove the events. Is this the desired behaviour?

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,297 @@
...
+		return errors.Wrap(err, "couldn't pop front oldest event from events deque")
+	}
+
+	var oldestCount octosql.Value
+
+	// Removing from events count map
+	err = eventsCountStorage.Get(&oldestEventTimeValue, &oldestCount)
+	if err != nil {
+		return errors.Wrap(err, "couldn't get oldest event count from events count map")
+	}
```
JasiekChomiak: Newline here?

execution/tvf/percentile_watermark_generator.go
cube2222: Please make this a float, otherwise a pretty standard percentile 99.9 is unrepresentable.

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,297 @@
...
+
+	return ws, execution.NewExecOutput(ws), nil // watermark generator stream now indicates new watermark source
+}
+
+// The way this watermark generator is working is following:
+// - events (must be positive) - represents amount of (most recent) events stored
+// - percentile (must be between 1 and 99) - represents percentile of recently stored events that are BIGGER than watermark value
+// 		ex. if percentile = 35 then 35% of events stored must be bigger than watermark value, so watermark position is at
+//			65th percentile of events stored (in sorted way)
+// - frequency (must be positive) - represents amount of events to be seen to initiate updating watermark process
```
cube2222: This comment line sounds like it starts after "frequency" records and later goes every one record. Maybe "represents amount of events to be seen before initiating next watermark update"?

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,297 @@
...
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't remove oldest event from storage")
+		}
+	}
+
+	// Updating watermark; enough events stored in deque AND frequency level reached
+	// '>=' for the case when we doesn't remove oldest event (explained above)
+	if eventsLength >= s.events && eventsSeen.AsInt() >= s.frequency {
+
+		// Clearing events seen
```
cube2222: More like resetting.

execution/tvf/percentile_watermark_generator.go
cube2222: Probably good to convert it to an integer later (of milliseconds? or microseconds?)

execution/tvf/percentile_watermark_generator.go
cube2222: Yes, I think it's ok.

execution/tvf/percentile_watermark_generator.go
woojtek: I personally think this would be an overkill, what about you, @cube2222 ?

execution/tvf/percentile_watermark_generator.go
woojtek: Quoting @cube2222 "Just a thing to keep in mind, watermarks can't decrease." :P

execution/tvf/percentile_watermark_generator.go
woojtek: By definition percentile can't actually be a float, only an integer between 1 and 99.
Because of that I don't think allowing float value in percentile expression would help in any way (then there would be int cast needed, <...>)

execution/tvf/percentile_watermark_generator.go
```diff
@@ -0,0 +1,300 @@
...
+	if eventsLength >= s.events && eventsSeen.AsInt() >= s.frequency {
+
+		// Resetting events seen
+		eventsSeen = octosql.MakeInt(0)
+
+		// Below declaration equals to (percentile*1000 / 100*1000) = (events - wP) / events
+		// Using (events - wP) instead of wP, because percentile of 80% means, that 80% events can be BIGGER than watermark value
+		// so watermark position is at 20% of all events
+		// Multiplying percentile by 1000 to allow float values like 99.9%
+		watermarkPosition := float64((100000-int(s.percentile*1000))*s.events) / 100000 // represents position (from the left) of event in sorted list that will become new watermark
```
cube2222: Why the float64 cast? The position always has to be a whole number, so an integer will suffice.

execution/tvf/percentile_watermark_generator.go
```diff
@@ -79,17 +79,17 @@ func (w *PercentileWatermarkGenerator) Get(ctx context.Context, variables octosq
 
 // The way this watermark generator is working is following:
 // - events (must be positive) - represents amount of (most recent) events stored
-// - percentile (must be between 1 and 99) - represents percentile of recently stored events that are BIGGER than watermark value
+// - percentile (must be positive and lesser than 100.0) - represents percentile of recently stored events that are BIGGER than watermark value
```
JasiekChomiak: less* than 100.0

PR #191, opened 2020-03-04
Deque.Length

woojtek:


streaming/storage/deque.go
```diff
@@ -265,6 +265,25 @@ func (dq *Deque) Clear() error {
 	return nil
 }
 
+func (dq *Deque) Length() (int, error) {
```
JasiekChomiak: Just a comment - since this calls both initializeBack() and initializeFront() it's not thread safe I think?

PR #192, opened 2020-03-09
Kafka support and pull engine batching 

cube2222:


execution/engine.go
```diff
@@ -82,53 +85,61 @@ func (engine *PullEngine) Run(ctx context.Context) {
...
+			engine.batchSizeManager.Reset()
 			tx = engine.storage.BeginTransaction()
 		} else if err != nil {
 			tx.Abort()
-			log.Println(err)
+			log.Println("engine: ", err)
 			return // TODO: Error propagation? Add this to the underlying queue as an ErrorElement? How to do this well? Send it to the underlying IRS like a watermark?
 		}
+
+		i++
```
woojtek: What is this `i` really for? Could You change it's name cause it's not clear for me :<

storage/kafka/datasource.go
```diff
@@ -0,0 +1,349 @@
...
+		DualStack: true,
+	}
+
+	r := kafka.NewReader(kafka.ReaderConfig{
+		Brokers:   ds.brokers,
+		Topic:     ds.topic,
+		Dialer:    dialer,
+		Partition: ds.partition,
+		MinBytes:  10e1,
+		MaxBytes:  10e7,
```
woojtek: Imho we could extract those values into some named constants

storage/kafka/datasource.go
```diff
@@ -0,0 +1,349 @@
...
+		decodeAsJSON: ds.decodeAsJSON,
+		alias:        ds.alias,
+	}
+	err := rs.loadOffset(tx)
+	if err != nil {
+		return nil, errors.Wrapf(err, "couldn't load kafka partition %d offset", ds.partition)
+	}
+
+	tokenQueue := execution.NewOutputQueue(tx.WithPrefix(tokenQueuePrefix))
+	for i := 0; i < 5; i++ {
```
woojtek: Why 5? Still, maybe it's better to make const `tokenCount` or something like that

storage/kafka/datasource.go
```diff
@@ -0,0 +1,349 @@
...
+	kafkaReader  *kafka.Reader
+	batchSize    int
+	decodeAsJSON bool
+	alias        string
+
+	workerCtxCancel    func()
+	workerCloseErrChan chan error
+}
+
+var outputQueuePrefix = []byte("$output_queue$")
```
woojtek: Can we keep all globals together so they are easier to find?

storage/kafka/datasource.go
```diff
@@ -0,0 +1,349 @@
...
+		values := []octosql.Value{octosql.MakeString(string(msg.Key))}
+		if !rs.decodeAsJSON {
+			fields = append(fields, octosql.NewVariableName(fmt.Sprintf("%s.value", rs.alias)))
+			values = append(values, octosql.MakeString(string(msg.Value)))
+		} else {
+			object := make(map[string]interface{})
+			err := json.Unmarshal(msg.Value, &object)
+			if err != nil {
+				log.Printf("couldn't decode value as json: %s, passing along as raw bytes", err)
+				fields = append(fields, octosql.NewVariableName("value"))
```
woojtek: Why there's no `rs.alias` here?

storage/kafka/datasource.go
```diff
@@ -0,0 +1,349 @@
...
+					values = append(values, octosql.NormalizeType(v))
+				}
+			}
+		}
+		batch[i] = execution.NewRecordFromSlice(fields, values)
+	}
+
+	for i := range batch {
+		err := outputQueue.Push(ctx, &QueueElement{
+			Type: &QueueElement_Record{
```
woojtek: Not very elegant way to create QueueElement :P - maybe you can add a constructor

storage/kafka/datasource.go
```diff
@@ -0,0 +1,349 @@
...
+
+	return nil
+}
+
+func (rs *RecordStream) loadOffset(tx storage.StateTransaction) error {
+	offsetState := storage.NewValueState(tx.WithPrefix(offsetPrefix))
+
+	var offset octosql.Value
+	err := offsetState.Get(&offset)
+	if err == storage.ErrNotFound {
```
woojtek: I don't know how offset works in kafka reader but shouldn't we save some 0 offset?

storage/kafka/datasource.go
```diff
@@ -0,0 +1,349 @@
...
+	offset := octosql.MakeInt(int(rs.kafkaReader.Offset()))
+	err := offsetState.Set(&offset)
+	if err != nil {
+		return errors.Wrap(err, "couldn't save kafka partition offset to state storage")
+	}
+
+	return nil
+}
+
+var readMessagesCountPrefix = []byte("$read_count$")
```
woojtek: Same with global position

execution/limit.go
```diff
@@ -100,12 +100,6 @@ func (node *LimitedStream) Next(ctx context.Context) (*Record, error) {
 	}
 
 	limit--
-	if limit == 0 {
```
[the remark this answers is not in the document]
cube2222: Because we can't close an underlying stream before we know that the transaction taking this record is commited. Had problems when testing.

config/getters.go
```diff
@@ -216,26 +240,47 @@ func GetIPAddress(config map[string]interface{}, field string, opts ...Option) (
 	return parts[0], int(port), nil
 }
 
-// GetStringList gets a string list from the given field.
-func GetStringList(config map[string]interface{}, field string, opts ...Option) ([]string, error) {
+// GetIPAddress gets an ip address from the given field.
+func GetIPAddressList(config map[string]interface{}, field string, opts ...Option) ([]string, []int, error) {
```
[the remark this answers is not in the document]
cube2222: comment added

execution/batch.go
```diff
@@ -0,0 +1,65 @@
...
+		latencyTarget: latencyTarget,
+		lastCommit:    time.Now(),
+		curBatchSize:  0,
+		batchSize:     10,
+	}
+}
+
+func (bsm *BatchSizeManager) CommitSuccessful() {
+	var delta int
+	delta = bsm.batchSize / 10
```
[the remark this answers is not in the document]
cube2222: yea, unnecessary here, result of many refactors.

execution/batch.go
```diff
@@ -0,0 +1,65 @@
...
+func NewBatchSizeManager(latencyTarget time.Duration) *BatchSizeManager {
+	return &BatchSizeManager{
+		latencyTarget: latencyTarget,
+		lastCommit:    time.Now(),
+		curBatchSize:  0,
+		batchSize:     10,
+	}
+}
+
+func (bsm *BatchSizeManager) CommitSuccessful() {
```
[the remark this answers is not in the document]
cube2222: comment added

execution/batch.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (bsm *BatchSizeManager) CommitAborted() {
+	bsm.curBatchSize = 0
+	bsm.lastCommit = time.Now()
+}
+
+func (bsm *BatchSizeManager) CommitTooBig() {
+	bsm.batchSize = bsm.batchSize / 2
+	if bsm.batchSize == 0 {
+		bsm.batchSize += 1
```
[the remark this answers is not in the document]
cube2222: sure

execution/engine.go
cube2222: relic of the past

storage/kafka/datasource.go
cube2222: I'm not for that. Because they're used in only this one location, with the descriptive variable next to them. It's like a const client config really. The variable name won't get any more descriptive. Those are some values I've chosen in the past because they worked for me, others would probably work well too.

storage/kafka/datasource.go
cube2222: Changed to constant. 5 seemed like a nice number :)

storage/kafka/datasource.go
cube2222: I'm putting them before the first function they are used. I can change this, but it's like that everywhere so far.

storage/kafka/datasource.go
cube2222: fixed

storage/kafka/datasource.go
cube2222: Why so? It's literally the only place this is used, and the structure is used as a raw package to satisfy the proto.Message protocol. I don't think there is anything to gain by making a constructor for this.

storage/kafka/datasource.go
cube2222: the kafka reader will default to something, but I decided to additionally add a startOffset config variable.

storage/kafka/datasource.go
cube2222: same answer

storage/kafka/datasource_test.go
```diff
@@ -0,0 +1,164 @@
+package kafka
```
JasiekChomiak: Isn't there anything else to test for this data source? I know it doesn't support any keys, but two test cases seems like a small amount :P

cube2222: No, not really.

execution/test_utils.go
```diff
@@ -45,6 +46,27 @@ func (rms *recordMultiSet) GetCount(rec *Record) int {
 	return 0
 }
 
+func (rms *recordMultiSet) isContained(other *recordMultiSet) bool {
+	for i, rec := range rms.set {
+		myCount := rms.count[i]
+		otherCount := other.GetCount(rec)
+		if otherCount != myCount {
```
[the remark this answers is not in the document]
woojtek: Yea that's exactly what I was thinking function named `isContained` is supposed to do

execution/test_utils.go
```diff
@@ -237,6 +276,55 @@ func ReadAll(ctx context.Context, stateStorage storage.Storage, stream RecordStr
...
+
+		rec, err := stream.Next(ctx)
+		if err == ErrEndOfStream {
+			err := tx.Commit()
+			if err != nil {
+				return nil, errors.Wrap(err, "couldn't commit transaction")
+			}
+			break
+		} else if errors.Cause(err) == ErrNewTransactionRequired {
+			err := tx.Commit()
```
woojtek: I guess this case could be rewritten to
`_ := tx.Commit() 
continue`

storage/kafka/datasource.go
```diff
@@ -0,0 +1,381 @@
...
+		streamID:     streamID,
+		kafkaReader:  r,
+		startOffset:  ds.startOffset,
+		batchSize:    ds.batchSize,
+		decodeAsJSON: ds.decodeAsJSON,
+		alias:        ds.alias,
+	}
+	err := rs.loadOffset(tx)
+	if err != nil {
+		return nil, execution.NewExecutionOutput(execution.NewZeroWatermarkGenerator()), errors.Wrapf(err, "couldn't load kafka partition %d offset", ds.partition)
```
woojtek: The concept in other datasources is return `nil` instead of `execution.NewExecutionOutput(execution.NewZeroWatermarkGenerator())` with early return due to error

storage/kafka/datasource.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	tokenQueue := execution.NewOutputQueue(tx.WithPrefix(tokenQueuePrefix))
+	for i := 0; i < initialTokenCount; i++ {
+		token := octosql.MakePhantom()
+		err := tokenQueue.Push(ctx, &token)
+		if err != nil {
+			return nil, execution.NewExecutionOutput(execution.NewZeroWatermarkGenerator()), errors.Wrap(err, "couldn't push batch token to queue")
```
woojtek: same

config/getters.go
```diff
@@ -134,6 +134,30 @@ func GetString(config map[string]interface{}, field string, opts ...Option) (str
 	return outString, nil
 }
 
+// GetStringList gets a string list from the given field.
+func GetStringList(config map[string]interface{}, field string, opts ...Option) ([]string, error) {
+	options := getOptions(opts...)
+	out, err := GetInterfaceList(config, field)
+	if err != nil {
+		if options.withDefault && errors.Cause(err) == ErrNotFound {
+			return options.defaultValue.([]string), nil
```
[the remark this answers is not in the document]
cube2222: We're doing it this way everywhere. This is called with constants. A user can't accidentally fail this. If this blows up, it's the programmers fault, so they should notice and fix it at development time.

config/getters.go
```diff
@@ -216,26 +240,47 @@ func GetIPAddress(config map[string]interface{}, field string, opts ...Option) (
...
-func GetStringList(config map[string]interface{}, field string, opts ...Option) ([]string, error) {
+// GetIPAddress gets an ip address list, as a list of hosts and a list of ports from the given field.
+func GetIPAddressList(config map[string]interface{}, field string, opts ...Option) ([]string, []int, error) {
 	options := getOptions(opts...)
 	out, err := GetInterfaceList(config, field)
 	if err != nil {
 		if options.withDefault && errors.Cause(err) == ErrNotFound {
-			return options.defaultValue.([]string), nil
+			defaults := options.defaultValue.([]interface{})
+			return defaults[0].([]string), defaults[1].([]int), nil
```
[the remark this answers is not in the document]
cube2222: same answer

execution/batch.go
```diff
@@ -0,0 +1,68 @@
...
+		batchSize:     10,
+	}
+}
+
+func (bsm *BatchSizeManager) CommitSuccessful() {
+	delta := bsm.batchSize / 10
+	if delta == 0 {
+		delta = 1
+	}
+	if time.Since(bsm.lastCommit) > bsm.latencyTarget {
```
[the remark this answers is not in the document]
cube2222: added

execution/engine.go
```diff
@@ -65,70 +66,77 @@ type PullEngine struct {
...
-		storage:         storage,
-		source:          source,
-		streamID:        streamID,
-		watermarkSource: watermarkSource,
+		irs:              irs,
+		storage:          storage,
+		source:           source,
+		streamID:         streamID,
+		watermarkSource:  watermarkSource,
+		batchSizeManager: NewBatchSizeManager(time.Second),
```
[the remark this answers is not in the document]
cube2222: I thought we could add this in the future when it's used in more places, to better decide on the ergonomics. For example, should we set the latency target for a single pull engine, or the whole plan combined?

execution/test_utils.go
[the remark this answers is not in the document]
cube2222: changed

storage/kafka/datasource.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+func NewDataSourceBuilderFactory(partitions int) physical.DataSourceBuilderFactory {
+	return physical.NewDataSourceBuilderFactory(
+		func(ctx context.Context, matCtx *physical.MaterializationContext, dbConfig map[string]interface{}, filter physical.Formula, alias string, partition int) (execution.Node, error) {
+			// Get execution configuration
+			hosts, ports, err := config.GetIPAddressList(dbConfig, "brokers", config.WithDefault([]interface{}{[]string{"localhost"}, []int{9092}}))
+			if err != nil {
+				return nil, errors.Wrap(err, "couldn't get address")
+			}
```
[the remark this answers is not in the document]
cube2222: I'm against this. It's always 4 lines per config param, so its really easy to parse, as there are no irregularities. Feel free to respond tho if you're really set on that approach.

storage/kafka/datasource.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+
+		fields := []octosql.VariableName{
+			octosql.NewVariableName(fmt.Sprintf("%s.key", rs.alias)),
+			octosql.NewVariableName(fmt.Sprintf("%s.offset", rs.alias)),
+		}
+		values := []octosql.Value{
+			octosql.MakeString(string(msg.Key)),
+			octosql.MakeInt(int(msg.Offset)),
+		}
```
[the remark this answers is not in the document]
cube2222: added

storage/kafka/datasource.go
```diff
@@ -0,0 +1,382 @@
...
+		if err == storage.ErrNotFound {
+			count = octosql.MakeInt(0)
+		} else if err != nil {
+			return nil, errors.Wrap(err, "couldn't read current read messages count")
+		}
+
+		newCount := count.AsInt() + 1
+
+		// If we're done with a batch, send back a token.
+		if newCount%rs.batchSize == 0 {
```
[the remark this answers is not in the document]
cube2222: gofmt wants no spaces here (if I add them it removes them)

execution/test_utils.go
cube2222: changed

storage/kafka/datasource.go
cube2222: oh yeah, sorry, changing

storage/kafka/datasource.go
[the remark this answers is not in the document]
JasiekChomiak: I'm not - I just have this approach that adding newlines seldom hurts readability, but you can leave it as is :)

PR #194, opened 2020-03-13
Distinct streaming

JasiekChomiak:
This PR introduces the "DISTINCT" functionality into our streaming world :)
The main issues that can still be thought about are:
1) Handling of repeating columns in output.go 
2) Forcing the determinisim of StarExpressions
3) Handling of outputFields in GroupBy


execution/group_by.go
```diff
@@ -259,3 +297,49 @@ func (gb *GroupByStream) Trigger(ctx context.Context, tx storage.StateTransactio
...
+	if !ok {
+		return 0
+	}
+
+	return count
+}
+
+func (f *groupByOutputFields) addField(field octosql.VariableName) {
+	f.outputFields = append(f.outputFields, field)
+	f.countMap[field] = f.getCount(field) + 1
```
woojtek: same as in variables file, you can just `+= 1` that element

execution/group_by.go
```diff
@@ -151,18 +162,45 @@ func (gb *GroupByStream) AddRecord(ctx context.Context, tx storage.StateTransact
...
+
+	recordVariables := record.AsVariables()
+
+	fieldCount := len(gb.inputFields)
+
+	values := make([]octosql.Value, fieldCount)
+
+	for i, fieldName := range gb.inputFields {
+		value := record.Value(fieldName)
+		values[i] = value
```
woojtek: Personally, I prefer to write that kind of code in 1 line only, more compact version doesn't reduce readability for me

execution/group_by.go
```diff
@@ -151,18 +162,45 @@ func (gb *GroupByStream) AddRecord(ctx context.Context, tx storage.StateTransact
...
+
+		outputFields = append(outputFields, starFields...)
+		values = append(values, starValues...)
+	}
+
+	gb.outputFields.extendByFields(outputFields)
+
+	// Add necessary aggregates
+	currentAggregateCount := len(gb.aggregates)
+	for i := 0; i < len(values)-currentAggregateCount; i++ {
```
woojtek: Might as well be `i := currentAggregateCount; i < len(values); i++`, could be a bit simpler

execution/execution.go
```diff
@@ -42,7 +42,50 @@ type Expression interface {
...
+// In ValueAndName, ExpressionValue and Name we ignore errors from Get
+// because we know that the fields we get are the ones that are surely
+// in the variables. Maybe we can panic here.
+func (se *StarExpression) ValueAndName(ctx context.Context, variables octosql.Variables) ([]octosql.Value, []octosql.VariableName, error) {
+	values := make([]octosql.Value, 0)
+	fields := make([]octosql.VariableName, 0)
+	keys := variables.DeterministicOrder()
+
+	for _, key := range keys {
+		if se.doesVariableMatch(key) {
```
cube2222: As we've been talking about, this won't work.

output/output.go
```diff
@@ -10,3 +10,34 @@ type Output interface {
...
+				currentRecordFieldMap[field] = count + 1
+			}
+
+			_, ok = fieldMap[field]
+			if !ok {
+				fieldMap[field] = 0
+			}
+		}
+
+		for _, field := range recordFields {
```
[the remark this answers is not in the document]
JasiekChomiak: Yeah you're right! I didn't know that you can do `map[key] += 1` and `if value > map[key]` even if the map has no mapping for `key` :D (I guess one learns something new everyday - thanks!)

JasiekChomiak:
Turns out this won't work at all :3

PR #196, opened 2020-03-15
Sql revival

woojtek:


storage/sql/template.go
```diff
@@ -68,7 +80,11 @@ func NewDataSourceBuilderFactoryFromTemplate(template SQLSourceTemplate) func([]
 
 				// Create a query with placeholders to prepare a statement from a physical formula
 				query := FormulaToSQL(filter, placeholders)
-				query = fmt.Sprintf("SELECT * FROM %s %s WHERE %s", tableName, alias, query)
+
+				// Adding Offset placeholder
+				offsetPlaceholder := placeholders.AddPlaceholder(physical.NewVariable(octosql.NewVariableName("offset")))
+
```
JasiekChomiak: I would randomize this somehow in case there is a variable called "offset" that appears in the Formula.
Or call this something like $offset_placeholder$

storage/sql/template.go
```diff
@@ -97,7 +115,30 @@ func NewDataSourceBuilderFactoryFromTemplate(template SQLSourceTemplate) func([]
...
+	rs := &RecordStream{
+		stateStorage: ds.stateStorage,
+		streamID:     streamID,
+		isDone:       false,
+		alias:        ds.alias,
+		batchSize:    ds.batchSize,
+	}
+	err := rs.loadOffset(tx)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "couldn't sql offset")
```
cube2222: couldn't _load_ sql offset

storage/sql/template.go
```diff
@@ -97,7 +115,30 @@ func NewDataSourceBuilderFactoryFromTemplate(template SQLSourceTemplate) func([]
...
+	}
+
+	// Adding offset value to variables
+	offsetVariable := octosql.NewVariables(map[octosql.VariableName]octosql.Value{"offset": octosql.MakeInt(rs.offset)})
```
JasiekChomiak: Since "offset" appears (at least) twice in this code, I would create some constant for this name

cube2222: change this to some magic name which the user is sure not to use maybe?

Overall I think we should change octosql.VariableName to provide a boolean if it's an internal variable, but this is definitely out of scope here :)

cube2222: Because offset is kinda normal

storage/sql/template.go
```diff
@@ -145,34 +199,167 @@ func (rs *RecordStream) Close() error {
...
+			err = waitableError.Close()
+			if err != nil {
+				log.Println("sql worker: couldn't close storage changes subscription: ", err)
+			}
+			continue
+		} else if err != nil {
+			tx.Abort()
+			log.Printf("sql worker: error running sql read batch worker: %s, reinitializing from storage", err)
+			tx := rs.stateStorage.BeginTransaction().WithPrefix(rs.streamID.AsPrefix())
+			if err := rs.loadOffset(tx); err != nil {
```
JasiekChomiak: We rarely do `if err := someFunction(); err != nil`. I think it's better to split in into the standard two lines:
`err := someFunction()`
`if err != nil ...`

cube2222: You have to create a new rs.rows here based on this offset. Overall I think it would be the cleanest if you just create two for loops (like I wrote you on messenger), where the outer creates the rows object for the current offset, and the inner just spins calling RunWorkerInternal.

woojtek: This is exactly how it is written in kafka, I was just sticking to it. But yea, can fix that :)

storage/sql/template.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+}
+
+var outputQueuePrefix = []byte("$output_queue$")
+
+func (rs *RecordStream) RunWorkerInternal(ctx context.Context, tx storage.StateTransaction) error {
+	outputQueue := execution.NewOutputQueue(tx.WithPrefix(outputQueuePrefix))
+
+	if rs.isDone {
+		err := outputQueue.Push(ctx, &kafka.QueueElement{
```
cube2222: You shouldn't have a dependency on the kafka package here. Create your own queue element.

PR #199, opened 2020-03-19
Remove reduntant maps optimization

JasiekChomiak:
This PR introduces a optimization scenario that reduces
Map{expressions: [], source: some_source, keep: true}
to
some_source,
since those two nodes yield the same results.

### DIFF
--- a/physical/optimizer/scenarios.go
+++ b/physical/optimizer/scenarios.go
@@ -12,11 +12,11 @@ import (
 
 var DefaultScenarios = []Scenario{
 	MergeRequalifiers,
-	RemoveEmptyMaps,
 	MergeFilters,
 	MergeDataSourceBuilderWithRequalifier,
 	MergeDataSourceBuilderWithFilter,
 	PushFilterBelowMap,
+	RemoveEmptyMaps,
 }
 
 var MergeRequalifiers = Scenario{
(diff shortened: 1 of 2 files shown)


physical/optimizer/scenarios.go
```diff
@@ -39,6 +40,31 @@ var MergeRequalifiers = Scenario{
...
+	CandidateMatcher: &MapMatcher{
+		Keep:        &AnyPrimitiveMatcher{Name: "keep"},
+		Expressions: &AnyNamedExpressionListMatcher{Name: "expressions"},
+		Source:      &AnyNodeMatcher{Name: "source"},
+	},
+
+	CandidateApprover: func(match *Match) bool {
+		expressions := match.NamedExpressionLists["expressions"]
+		keep := match.Primitives["keep"]
+		keepCast, ok := keep.(bool)
```
cube2222: I don't think you have to check the ok, just do
```
keepCast := keep.(bool)
```
if it's not a bool, then there's a serious programmatic error, and it shouldn't just return false.

physical/optimizer/scenarios.go
```diff
@@ -12,6 +12,7 @@ import (
 
 var DefaultScenarios = []Scenario{
 	MergeRequalifiers,
+	RemoveEmptyMaps,
```
cube2222: Please keep _some_ _semblance_ of order here. And don't stick a Remove between two Merge'es :P

PR #200, opened 2020-03-20
Star Expressions

JasiekChomiak:
This PR introduces the StarExpression that allow users to write queries like
"SELECT d.* FROM ..."

### DIFF
--- a/execution/execution.go
+++ b/execution/execution.go
@@ -2,6 +2,7 @@ package execution
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/cube2222/octosql"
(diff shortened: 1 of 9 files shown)


execution/map.go
```diff
@@ -64,21 +64,39 @@ func (stream *MappedStream) Next(ctx context.Context) (*Record, error) {
...
+			for _, name := range expressionName {
+				value, _ := variables.Get(name) //TODO: err is always nil, handle it anyways?
+
+				outValues[name] = value
+			}
 
-		value, err := expr.ExpressionValue(ctx, variables)
-		if err != nil {
-			return nil, errors.Wrapf(err, "couldn't get expression %v", expr.Name())
+			fieldNames = append(fieldNames, expressionName...)
```
woojtek: In `default` case you first append `fieldNames` and then set `outValues` so I think you could move this line above for loop

execution/execution.go
```diff
@@ -45,6 +46,54 @@ type NamedExpression interface {
...
+	return fields
+}
+
+func (se *StarExpression) ExpressionValue(ctx context.Context, variables octosql.Variables) (octosql.Value, error) {
+	fields := se.Fields(variables)
+	values := make([]octosql.Value, len(fields))
+
+	for i := range fields {
+		value, err := variables.Get(fields[i])
+		if err != nil { // this shouldn't happen, but we might as well check
```
cube2222: if it shouldn't, then maybe panic

execution/map.go
```diff
@@ -64,21 +64,38 @@ func (stream *MappedStream) Next(ctx context.Context) (*Record, error) {
...
 	for _, expr := range stream.expressions {
-		fieldNames = append(fieldNames, expr.Name())
+		switch expr := expr.(type) {
+		case *StarExpression:
+			// A star expression is based only on the incoming record
+			expressionName := expr.Fields(recordVariables)
+			fieldNames = append(fieldNames, expressionName...)
+
+			for _, name := range expressionName {
+				value, _ := variables.Get(name) //TODO: err is always nil, handle it anyways?
```
cube2222: yea, feel free to panic

parser/parser.go
```diff
@@ -192,26 +217,52 @@ func ParseSelect(statement *sqlparser.Select) (logical.Node, error) {
...
+				}
+			}
 		}
+
+		// Handle star expressions collisions
+		for i := len(nonStarExpressions); i < len(statement.SelectExprs); i++ {
+			qualifier := expressions[i].Name().Source()
+
+			if qualifier == "" { // a * expression collides with anything else
+				if len(statement.SelectExprs) > 1 {
```
cube2222: no! it doesn't actually.
this is bad: SELECT *, p.name FROM people p
this isn't: SELECT *, p.name as mename FROM people p
or even: SELECT test, (SELECT *, p.test FROM people g) FROM people p

Though we can ignore that for now and leave a TODO here, as you can easily work around it using qualified star expressions.

parser/parser.go
```diff
@@ -428,7 +479,7 @@ func ParseAggregate(expr sqlparser.Expr) (logical.Aggregate, logical.NamedExpres
 				return "", nil, errors.Wrap(err, "couldn't parse aggregate argument")
 			}
 
-		case *sqlparser.StarExpr:
+		case *sqlparser.StarExpr: //TODO: not sure this is reachable tbh
```
cube2222: Isn't it the case for COUNT(*) or are they handled differently?

At least panic, and don't leave a mine like parseArg = nil (I think it's a mine, isn't it?)

parser/parser.go
JasiekChomiak: Oh you are absolutely right! But to be honest I don't see a simple fix for this rn, because at the time of parsing the query you have no way of telling what the column names are gonna be. Leaving as TODO

parser/parser.go
JasiekChomiak: Oh it actually is reachable, and the nil is because when you return from it, if the result is nil you know you are aggregating stars

PR #201, opened 2020-03-20
Distinct streaming II - The Distinct Strikes Back

JasiekChomiak:
This PR introduces Distinct into our streaming world (again...) :D
No tests yet.
This is the first draft of the implementation.


execution/distinct.go
```diff
@@ -24,104 +28,162 @@ func (node *Distinct) Get(ctx context.Context, variables octosql.Variables, stre
...
 	}
 
-	stream, execOutput, err := node.child.Get(ctx, variables, sourceStreamID)
+	source, execOutput, err := node.source.Get(ctx, variables, sourceStreamID)
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "couldn't get stream for child node in distinct")
+		return nil, nil, errors.Wrap(err, "couldn't get stream for source node in distinct")
+	}
+
+	trigger, err := NewCountingTrigger(NewDummyValue(octosql.MakeInt(1))).Get(ctx, variables)
```
cube2222: Please write a comment on why it's the dummy value integer 1.

I know it, you know it, but somebody else might not know it.

execution/distinct.go
```diff
@@ -24,104 +28,162 @@ func (node *Distinct) Get(ctx context.Context, variables octosql.Variables, stre
...
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "couldn't create trigger for distinct")
+	}
+
+	distinct := &DistinctStream{
+		variables: variables,
+	}
+
+	processFunc := &ProcessByKey{
+		eventTimeField:  "", // TODO: what about this?
```
cube2222: You have to handle it similarly as in group by. Using the event time for distinct is important, as it allows distinct usage on streams.

execution/distinct.go
```diff
@@ -24,104 +28,162 @@ func (node *Distinct) Get(ctx context.Context, variables octosql.Variables, stre
...
+	return nil, nil
 }
 
-func (rs *recordSet) Insert(r *Record) (bool, error) {
-	hash, err := r.Hash()
-	if err != nil {
-		return false, errors.Wrap(err, "couldn't get hash of record")
-	}
+// TODO: move this somewhere else?
+func recordFromMap(m map[string]octosql.Value, isUndo bool) *Record {
```
cube2222: This doesn't handle record event times.

execution/distinct.go
```diff
@@ -24,104 +28,162 @@ func (node *Distinct) Get(ctx context.Context, variables octosql.Variables, stre
...
+	if !wasRecordTriggered && isRecordPresent {
+		// This is the first time that record is triggered and it is present, so we send it
+
+		// Save that the record was triggered
+		err = recordTriggerState.Set(&phantom)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't mark record as retracted")
+		}
+
+		record := recordFromMap(recordMap, false)
```
cube2222: Maybe just store the record somewhere instead of reconstructing it from the key.

This way we will surely not lose stuff like metadata, event time, an ID etc.

execution/distinct.go
```diff
@@ -24,104 +30,219 @@ func (node *Distinct) Get(ctx context.Context, variables octosql.Variables, stre
...
+		panic("no id for record in distinct")
 	}
 
-	return nil
-}
+	keyPrefix := append(append([]byte("$"), key.MonotonicMarshal()...), '$')
+	txByKey := tx.WithPrefix(keyPrefix)
+
+	// Keep track of record vs retraction count
+	recordCountState := storage.NewValueState(txByKey.WithPrefix(recordCountPrefix))
```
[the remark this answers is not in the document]
JasiekChomiak: It's actually defined in Group By and I can't allowed it locally. I can always rename it to `distinctRecordCountPrefix` and define it in the distinct.go file

PR #202, opened 2020-03-21
Csv/Json/Excel revival

woojtek:
Alright, so few cases I didn't know exactly how to deal with (also, most of them are marked as TODOs in code):
- Firstly, what Janek pointed out, during actual storage revival (and moving offset) do we have to do anything more sophisticated than moving file iterator one by one? This might be a little bit slow but all of these packages doesn't seem to have any seek-like method implemented, so that would be a bit complex (moving file iterators on our own by remembering bytes read, bleh)
- CSV/Excel: If the above point is fine, are `readRecordFromFileWithInitialize()` methods acceptable, I mean, to make it quite simple rather than just read raw line from file these functions also parse records, but thanks to that we can also handle header rows and stuff like that, so the code is just much simpler and easier to understand (although, when moving by offset we actually parse records and drop them afterwards)
- JSON/CSV/Excel: when couldn't read line, error is returned, but there are few ways to handle this:
1. Drop the entire batch and instantly return error -> currently implemented (but pretty silly imho)
2. Push first records that are correct (part of batch) then return error -> quite easy to do
3. Maybe even don't bother and continue reading like this incorrect record didn't appear...
...
- ... What should happen to worker after file read error occurs? Currently (copy pasted from kafka) we just reenter outer loop so we actually continue reading the rest of the file (example: let's say we have 3 rows where 2nd one is somehow incorrect and batchSize = 2. we will drop the first full batch and then continue reading the file: output -> 3rd row). I believe we should distinguish file errors (which instantly closes worker) from standard not-nil errors like queue push errors or badger errors (which reinitialize reding from storage).


storage/csv/datasource.go
```diff
@@ -112,13 +139,163 @@ func (rs *RecordStream) Close() error {
...
+		if err != nil {
+			log.Fatalf("csv worker: couldn't reinitialize offset for csv read batch worker: %s", err)
+			return
+		}
+
+		tx.Abort() // We only read data above, no need to risk failing now.
+
+		// Moving file iterator by `rs.offset`
+		for i := 0; i < rs.offset; i++ {
+			_, err := rs.readRecordFromFileWithInitialize()
```
JasiekChomiak: Can't this be done more efficiently? I mean move the offset all at once?
Like initialize first and then just move by offset lines?

storage/json/datasource.go
```diff
@@ -95,11 +122,72 @@ func (rs *RecordStream) Close() error {
...
+		err := rs.loadOffset(tx)
+		if err != nil {
+			log.Fatalf("json worker: couldn't reinitialize offset for json read batch worker: %s", err)
+			return
+		}
+
+		tx.Abort() // We only read data above, no need to risk failing now.
+
+		// Moving file iterator by `rs.offset`
+		for i := 0; i < rs.offset; i++ {
```
JasiekChomiak: The same applies as in CSV.

storage/csv/datasource.go
woojtek: I'm afraid not with a use of those packages, the way to handle this better (possibly) is to store bytes read after every line and then use Seek on the file itself but it's pretty disgusting imho :(

cube2222:
So, just to start with. Regarding the intro you wrote:
- Seeking one by one is ok imo
- You definitely don't want to just skip on error. Kafka reinitializes the offset, so will try to read the batch again. Failing the whole batch is ok. Just retry it afterwards, as in kafka. Reset the offset basically. And log something.

execution/record.go
```diff
@@ -179,7 +179,7 @@ func (r *Record) Show() string {
 		parts[i] = fmt.Sprintf("%s: %s", field.Name, r.Value(field.Name).Show())
 	}
 
-	return fmt.Sprintf("{%s}", strings.Join(parts, ", "))
+	return fmt.Sprintf("{%s, %s}", r.Metadata.Id, strings.Join(parts, ", "))
```
cube2222: If you add the record ID to the show, at least add it in a form as above. Make r.ShowFields() return sys.id, and handle this in Value() to return the id.

cube2222: Don't special case it.

storage/csv/datasource.go
cube2222: Yea, don't optimize it for now.

storage/csv/datasource.go
```diff
@@ -112,13 +139,161 @@ func (rs *RecordStream) Close() error {
...
+				}
+				continue
+			} else if err == execution.ErrEndOfStream {
+				err = tx.Commit()
+				if err != nil {
+					log.Println("csv worker: couldn't commit transaction: ", err)
+					continue
+				}
+				return
+			} else if err != nil { // TODO - should this close worker? This could be a case where file is somehow faulty (question in PR)
```
cube2222: Answered in PR.

storage/csv/datasource.go
```diff
@@ -145,66 +320,86 @@ func (rs *RecordStream) initializeColumnsWithHeaderRow() error {
...
-	return execution.NewRecord(
-		rs.aliasedFields,
-		aliasedRecord,
-		execution.WithID(execution.NewRecordIDFromStreamIDWithOffset(rs.streamID, curOffset)),
-	), nil
+	switch queueElement := queueElement.Type.(type) {
+	case *QueueElement_Record:
+		return queueElement.Record, nil
+	case *QueueElement_Error:
+		if queueElement.Error == execution.ErrEndOfStream.Error() {
```
cube2222: Don't compare strings. Add a message type EndOfStream and send it. Same as in join or ProcessByKey.

storage/csv/datasource.go
```diff
@@ -145,66 +320,86 @@ func (rs *RecordStream) initializeColumnsWithHeaderRow() error {
...
-	}
+	rs.offset = offset.AsInt()
+
+	return nil
+}
 
+func (rs *RecordStream) saveOffset(tx storage.StateTransaction, curBatchSize int) error {
+	offsetState := storage.NewValueState(tx.WithPrefix(offsetPrefix))
+
+	rs.offset = rs.offset + curBatchSize
```
cube2222: I'm not sure if saveOffset should include this addition. That's pretty unexpected. Maybe move it out of here and then call saveOffset after adding the current batch size.

storage/excel/datasource.go
```diff
@@ -135,22 +153,134 @@ type RecordStream struct {
...
+var outputQueuePrefix = []byte("$output_queue$")
+
+func (rs *RecordStream) RunWorkerInternal(ctx context.Context, tx storage.StateTransaction) error {
+	outputQueue := execution.NewOutputQueue(tx.WithPrefix(outputQueuePrefix))
+
 	if rs.isDone {
-		return nil, execution.ErrEndOfStream
+		err := outputQueue.Push(ctx, &QueueElement{
+			Type: &QueueElement_Error{
+				Error: execution.ErrEndOfStream.Error(),
```
cube2222: as above

storage/json/datasource.go
```diff
@@ -95,11 +122,144 @@ func (rs *RecordStream) Close() error {
...
+var outputQueuePrefix = []byte("$output_queue$")
+
+func (rs *RecordStream) RunWorkerInternal(ctx context.Context, tx storage.StateTransaction) error {
+	outputQueue := execution.NewOutputQueue(tx.WithPrefix(outputQueuePrefix))
+
 	if rs.isDone {
-		return nil, execution.ErrEndOfStream
+		err := outputQueue.Push(ctx, &QueueElement{
+			Type: &QueueElement_Error{
+				Error: execution.ErrEndOfStream.Error(),
```
cube2222: same

storage/csv/datasource.go
```diff
@@ -112,13 +139,161 @@ func (rs *RecordStream) Close() error {
...
+		}
+
+		batch = append(batch, execution.NewRecord(
+			rs.aliasedFields,
+			aliasedRecord,
+			execution.WithID(execution.NewRecordIDFromStreamIDWithOffset(rs.streamID, rs.offset+i))))
+	}
+
+	for i := range batch {
+		err := outputQueue.Push(ctx, &QueueElement{
```
JasiekChomiak: This applies to all places this appears - I think it would be more optimal to write a `NewQueueRecordElement` or something of that sort and use it everywhere, because 4 lines of code is a bit of an overkill to create the queue element, and this appears more than once throughout the code

storage/excel/datasource.go
```diff
@@ -191,17 +322,74 @@ func (rs *RecordStream) Next(context.Context) (*execution.Record, error) {
...
+	} else if err != nil {
+		return errors.Wrap(err, "couldn't load excel offset from state storage")
+	}
+
+	rs.offset = offset.AsInt()
+
+	return nil
+}
+
+func (rs *RecordStream) saveOffset(tx storage.StateTransaction, curBatchSize int) error {
```
JasiekChomiak: Here the same thing as above applies as well (what @cube2222 mentioned)

storage/json/datasource.go
JasiekChomiak: Feel free to resolve this convo btw ;)

storage/csv/datasource.go
woojtek: I wrote about that in cube's kafka PR (I mean, exactly the same question) and we've come into conclusion that it wouldn't be necessary so I copy-pasted the way kafka datasource does it
@cube2222

storage/csv/datasource.go
woojtek: moved `rs.offset = rs.offset + curBatchSize` before `saveOffset` call

storage/csv/datasource.go
```diff
@@ -112,13 +139,161 @@ func (rs *RecordStream) Close() error {
...
+		if err == execution.ErrEndOfStream {
+			break
+		} else if err != nil { // TODO - do we want to return err instantly here or push current batch? (question in PR)
+			return errors.Wrap(err, "couldn't read record from csv file")
+		}
+
+		batch = append(batch, execution.NewRecord(
+			rs.aliasedFields,
+			aliasedRecord,
+			execution.WithID(execution.NewRecordIDFromStreamIDWithOffset(rs.streamID, rs.offset+i))))
```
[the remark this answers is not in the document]
woojtek: I don't think so :/

storage/csv/datasource.go
cube2222: yea, don't abstract too early. I actually think that we'll be able to abstract out most of the "revival" logical in the future.

storage/csv/datasource.go
```diff
@@ -112,13 +128,174 @@ func (rs *RecordStream) Close() error {
...
+		err := outputQueue.Push(ctx, &QueueElement{
+			Type: &QueueElement_EndOfStream{
+				EndOfStream: true,
+			},
+		})
+		if err != nil {
+			return errors.Wrapf(err, "couldn't push csv EndOfStream to output record queue")
+		}
+
+		log.Println("csv worker: ErrEndOfStream pushed")
```
cube2222: remove debug log before merging

storage/csv/datasource.go
```diff
@@ -112,13 +128,174 @@ func (rs *RecordStream) Close() error {
...
+		err := outputQueue.Push(ctx, &QueueElement{
+			Type: &QueueElement_Record{
+				Record: batch[i],
+			},
+		})
+		if err != nil {
+			return errors.Wrapf(err, "couldn't push csv record with index %d in batch to output record queue", i)
+		}
+
+		log.Println("csv worker: record pushed: ", batch[i])
```
cube2222: same

Issue #203, opened 2020-03-23
I like octosql, I would gladly contribute my efforts，I want to complete Visualizer with graphviz


runner-mei:
- [x] 1.  all Node type of logical.  #204
- [x] 2.  ~~all Node type of physical.~~
- [x] 2. add visualize cmd #205

cube2222:
Great to hear that! 🙂

Are there any types of physical nodes which are currently not visualized?

Anyways, feel free to work on this, nobody is currently working on it afaik, and feel free to open an early WIP pull request, so we can stay connected while you work on it!

runner-mei:
https://github.com/runner-mei/octosql/commit/638fece867b71f067898cfd8097b308ead99bb9c

PR #204, opened 2020-03-25
feat(logical): add logical visualizer

runner-mei:
fix  https://github.com/cube2222/octosql/issues/203

PR #205, opened 2020-03-25
feat(cmd): add sqlviz command

runner-mei:
fix https://github.com/cube2222/octosql/issues/203
include https://github.com/cube2222/octosql/issues/204


runner-mei:
I agree with your opinion, I just don't want to write configuration and add it.

PR #207, opened 2020-03-25
Redis revival

woojtek:


storage/redis/datasource.go
```diff
@@ -146,91 +159,312 @@ func (rs *KeySpecificStream) Close() error {
...
+var outputQueuePrefix = []byte("$output_queue$")
+
+func (rs *RecordStream) RunWorkerInternal(ctx context.Context, tx storage.StateTransaction) error {
+	outputQueue := execution.NewOutputQueue(tx.WithPrefix(outputQueuePrefix))
+
 	if rs.isDone {
-		return nil, execution.ErrEndOfStream
+		err := outputQueue.Push(ctx, &QueueElement{
+			Type: &QueueElement_Error{
+				Error: execution.ErrEndOfStream.Error(),
```
cube2222: same

storage/redis/datasource.go
```diff
@@ -146,91 +159,312 @@ func (rs *KeySpecificStream) Close() error {
...
+	batch := make([]*execution.Record, 0)
+	for i := 0; i < len(allKeysForThisBatch); i++ {
+		key := allKeysForThisBatch[i]
+		recordValues, err := rs.client.HGetAll(key).Result()
+		if err != nil {
+			return errors.Wrapf(err, "could't get hash for key %s", key)
+		}
+
+		// We skip this record
+		if len(recordValues) == 0 {
```
cube2222: Can this happen? What does this mean? Is it same as not found or can the value be empty. If it can be empty, the just return a record with only the key.

woojtek: explained irl

storage/redis/datasource.go
```diff
@@ -93,51 +103,54 @@ func (ds *DataSource) Get(ctx context.Context, variables octosql.Variables, stre
...
+		batchSize:    ds.batchSize,
 	}
 
-	sliceKeys := make([]string, 0)
-	for k := range keysWanted.keys {
-		sliceKeys = append(sliceKeys, k)
+	if len(keysWanted.keys) == 0 { // EntireDatabaseStream
+		rs.isEntireDatabaseStream = true
+	} else { // KeySpecificStream
+		sliceKeys := make([]string, 0)
```
JasiekChomiak, line 118: I know this wasn't introduced in this PR, but it just struck me. Isn't it better to simply write 
`rs.keys = make([]string, 0)`
`for k := range keysWanted.keys {`
     `    rs.keys = append(rs.keys, k)`
`}`

woojtek: yea you're right, after refactors this looks kinda dumb :P

PR #209, opened 2020-03-31
Badger std output

cube2222:


output/badger/stdout.go
```diff
@@ -0,0 +1,103 @@
...
+	for range time.Tick(time.Second / 10) {
+		tx := printer.stateStorage.BeginTransaction()
+
+		var buf bytes.Buffer
+
+		records, err := printer.recordsLister.ListRecords(ctx, tx)
+		if err != nil {
+			return errors.Wrap(err, "couldn't list records")
+		}
+		if len(records) > 0 && !records[0].EventTimeField().Empty() {
```
woojtek, line 47: maybe newline here

output/badger/output.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	IDs := octoIDs.AsSlice()
+
+	if !record.IsUndo() {
+		newIDs := octosql.MakeTuple(append(IDs, octosql.MakeString(record.ID().Show())))
+		if err := records.Set(&key, &newIDs); err != nil {
+			return errors.Wrap(err, "couldn't add record ID to output records")
+		}
+	} else {
+		IDs = IDs[:len(IDs)-1]
```
woojtek, line 56: I don't think I understand, how do You know that ID of current record (with undo) is the last one in IDs slice?

output/badger/output.go
```diff
@@ -0,0 +1,202 @@
...
+					execution.WithID(execution.NewRecordID(id.AsString())),
+					execution.WithEventTimeField(o.EventTimeField),
+				),
+			)
+		}
+	}
+	if err := iter.Close(); err != nil {
+		return nil, errors.Wrap(err, "couldn't close iterator")
+	}
+	if err != storage.ErrEndOfIterator {
```
woojtek, line 197: Shouldn't this be directly after for loop? It probably doesn't matter but conceptually first we want to check if the for loop finished without an error and then close the iterator

output/badger/output.go
cube2222: undo's have different ID's than the original records. So we just delete any one of them.

output/badger/output.go
cube2222: sure

PR #214, opened 2020-04-05
Close cleanup

woojtek:


execution/engine.go
```diff
@@ -210,16 +210,20 @@ func (engine *PullEngine) GetWatermark(ctx context.Context, tx storage.StateTran
...
 		return errors.Wrap(err, "couldn't close source stream")
 	}
 
-	err = engine.irs.Close()
-	if err != nil {
+	if err := engine.irs.Close(ctx); err != nil {
 		return errors.Wrap(err, "couldn't close intermediate record store")
 	}
 
+	// No badger writes are made by pull engine but they are by irs
```
cube2222, line 222: This comment may shortly become invalid, so please just leave it out.

execution/process.go
```diff
@@ -223,6 +223,11 @@ func (p *ProcessByKey) ReadyForMore(ctx context.Context, tx storage.StateTransac
 	return nil
 }
 
-func (p *ProcessByKey) Close() error {
-	return nil // TODO: Close this, remove state
+func (p *ProcessByKey) Close(ctx context.Context) error {
+	storage := storage.GetStateTransactionFromContext(ctx).GetUnderlyingStorage()
+	if err := storage.DropAll(outputWatermarkPrefix); err != nil {
```
cube2222, line 228: This is actually prefixed by the stream ID. (the Next() and GetWatermark() methods get a tx prefixed by the StreamID)

cube2222: So the cleanup in the Pull engine is sufficient.

execution/tvf/percentile_watermark_generator.go
```diff
@@ -130,10 +132,10 @@ func (s *PercentileWatermarkGeneratorStream) Next(ctx context.Context) (*executi
 	}
 
 	tx := storage.GetStateTransactionFromContext(ctx)
-	watermarkStorage := storage.NewValueState(tx.WithPrefix(percentileWatermarkPrefix))
-	eventsStorage := storage.NewDeque(tx.WithPrefix(percentileWatermarkEventsPrefix))
-	eventsCountStorage := storage.NewMap(tx.WithPrefix(percentileWatermarkEventsCountPrefix))
-	eventsSeenStorage := storage.NewValueState(tx.WithPrefix(percentileWatermarkEventsSeenPrefix))
+	watermarkStorage := storage.NewValueState(tx.WithPrefix(s.streamID.AsPrefix()).WithPrefix(percentileWatermarkPrefix))
```
cube2222, line 135: Please just create one tx with the prefix added.

tx = tx.WithPrefix(...)

execution/tvf/watermark_generator.go
```diff
@@ -106,7 +108,7 @@ func (s *WatermarkGeneratorStream) Next(ctx context.Context) (*execution.Record,
 	}
 
 	if timeValueWithOffset.After(currentWatermark) { // time in current record is bigger than current watermark - update it
-		watermarkStorage := storage.NewValueState(tx.WithPrefix(watermarkPrefix))
+		watermarkStorage := storage.NewValueState(tx.WithPrefix(s.streamID.AsPrefix()).WithPrefix(watermarkPrefix))
```
cube2222, line 111: as above

storage/csv/datasource.go
```diff
@@ -119,12 +119,16 @@ type RecordStream struct {
 	batchSize       int
 }
 
-func (rs *RecordStream) Close() error {
-	err := rs.file.Close()
-	if err != nil {
+func (rs *RecordStream) Close(ctx context.Context) error {
```
cube2222, line 122: I think this won't shut off the worker?

cube2222: I think it would be better if there was a boolean variable $running$ in badger. It would be set to false. Then, when the worker noticed it, it would shut itself down and delete the underlying storage.

cube2222: Ah no, Close is not transactional because of DropAll.

cube2222: So we can just cancel some context.

cube2222: Like with the subscriptions.

execution/tvf/watermark_generator.go
```diff
@@ -118,6 +120,15 @@ func (s *WatermarkGeneratorStream) Next(ctx context.Context) (*execution.Record,
...
 }
 
-func (s *WatermarkGeneratorStream) Close() error {
-	return s.source.Close()
+func (s *WatermarkGeneratorStream) Close(ctx context.Context) error {
+	if err := s.source.Close(ctx); err != nil {
+		return errors.Wrap(err, "couldn't close underlying stream")
+	}
+
+	storage := storage.GetStateTransactionFromContext(ctx).GetUnderlyingStorage()
```
cube2222, line 128: Close isn't transactional, so there might not be a transaction here.

cube2222: Maybe the Close context should contain the Storage itself? Or Close should have a signature Close(ctx context.Context, storage storage.Storage) so that everybody gets it?

woojtek: I think expanding the signature by `storage.Storage` is a good and relatively easy to implement idea

app/app.go
```diff
@@ -68,11 +68,10 @@ func (app *App) RunPlan(ctx context.Context, stateStorage storage.Storage, plan
 
 	outStreamID := &execution.StreamID{Id: "output"}
 
-	pullEngine := execution.NewPullEngine(out, stateStorage, stream[0], outStreamID, execOutput[0].WatermarkSource, true)
+	ctx, cancel := context.WithCancel(ctx)
+	pullEngine := execution.NewPullEngine(out, stateStorage, stream[0], outStreamID, execOutput[0].WatermarkSource, true, cancel)
```
cube2222, line 72: Give the pull engine a context, then do context.WithCancel(ctx) in there. And then check ctx.Done on it. (you can remove the ctx from pullEngine.Run())

storage/json/datasource.go
```diff
@@ -101,33 +114,50 @@ type RecordStream struct {
...
 
 	return nil
 }
 
-func (rs *RecordStream) RunWorker(ctx context.Context) {
+func (rs *RecordStream) RunWorker(ctx context.Context) error {
 	for { // outer for is loading offset value and moving file iterator
+		select {
+		case <-ctx.Done():
+			return nil
```
cube2222, line 145: Usually you'd just return the error here ctx.Err(), and just handle the Canceled and DeadlineExceeded errors above.

storage/json/datasource.go
```diff
@@ -80,11 +80,24 @@ func (ds *DataSource) Get(ctx context.Context, variables octosql.Variables, stre
 		batchSize:                     ds.batchSize,
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
+	rs.workerCtxCancel = cancel
+	rs.workerCloseErrChan = make(chan error)
```
cube2222, line 85: It's better to make this channel buffered with capacity of 1. This way if the worker errors and there's nobody to read the channel, the goroutine won't hang on channel send.

storage/json/datasource.go
```diff
@@ -101,33 +115,50 @@ type RecordStream struct {
...
 }
 
-func (rs *RecordStream) Close() error {
-	err := rs.file.Close()
-	if err != nil {
-		return errors.Wrap(err, "Couldn't close underlying file")
+func (rs *RecordStream) Close(ctx context.Context, storage storage.Storage) error {
+	rs.workerCtxCancel()
+	err := <-rs.workerCloseErrChan
+	if err == context.Canceled {
```
cube2222, line 126: check for deadline exceeded too

execution/engine.go
```diff
@@ -64,68 +64,96 @@ type PullEngine struct {
...
 			tx = engine.storage.BeginTransaction()
 		} else if err != nil {
-			tx.Abort()
 			log.Println("engine: ", err)
+			tx.Abort()
 			tx = engine.storage.BeginTransaction()
-			err := engine.irs.MarkError(ctx, tx.WithPrefix(engine.streamID.AsPrefix()), err)
-			if err != nil {
-				log.Fatalf("couldn't mark error on intermediate record store: %s", err)
+			err1 := engine.irs.MarkError(engine.ctx, engine.getPrefixedTx(tx), err)
```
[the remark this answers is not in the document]
woojtek: I think storing old error value might be a bit better - introducing `markErrorErr` would mean introducing also sth like `commitErr` which I don't really like

PR #221, opened 2020-04-23
Fix stream branches with shuffling

cube2222:


execution/lookup_join.go
```diff
@@ -250,24 +250,24 @@ func (rs *LookupJoinStream) RunWorker(ctx context.Context, id *RecordID) error {
...
 
-	if err := tx.Commit(); err != nil {
-		return errors.Wrapf(err, "couldn't commit transaction setting up the job")
-	}
-
 	engine := NewPullEngine(
 		&JobOutputQueueIntermediateRecordStore{recordID: id},
 		rs.stateStorage,
-		joinedStream,
+		joinedStream[0],
```
JasiekChomiak, line 270: Why do you only take the first stream? Will only one be created? If so, add a comment saying that

PR #222, opened 2020-04-29
Streaming Output and Adding Back Output Formats

cube2222:


app/app.go
```diff
@@ -20,11 +20,11 @@ import (
 type App struct {
 	cfg                  *config.Config
 	dataSourceRepository *physical.DataSourceRepository
-	out                  output.Output
+	out                  interface{}
```
JasiekChomiak, line 23: Explain this please, `interface{}` is the least self-explanatory thing there is, so it deserves a comment

cmd/octosql/main.go
```diff
@@ -77,23 +81,52 @@ With OctoSQL you don't need O(n) client tools or a large data analysis system de
 			log.Fatal(err)
 		}
 
-		var out output.Output
+		// stream json
```
JasiekChomiak, line 84: What is this comment for?

cmd/octosql/main.go
```diff
@@ -77,23 +81,52 @@ With OctoSQL you don't need O(n) client tools or a large data analysis system de
...
+				sink := batch.NewTableOutput(streamID, eventTimeField)
+				out := batch.NewWholeTablePrinter(stateStorage, sink, batchtable.TableFormatter(false))
+				return sink, out
+			}
 		default:
 			log.Fatal("invalid output type")
 		}
 
-		app := app.NewApp(cfg, dataSourceRespository, out, describe)
+		app := app.NewApp(cfg, dataSourceRespository, nil, describe)
```
JasiekChomiak, line 129: What? Why is this `nil`?

cmd/octosql/main.go
```diff
@@ -77,23 +81,52 @@ With OctoSQL you don't need O(n) client tools or a large data analysis system de
...
 		}
 
-		var out output.Output
+		// stream json
+		// live csv
+		// live table
+		// batch csv
+		// batch table
+		var outputSinkFn func(stateStorage storage.Storage, streamID *execution.StreamID, eventTimeField octosql.VariableName) (execution.IntermediateRecordStore, output.Printer)
 		switch outputFormat {
```
JasiekChomiak, line 90: OK to be honest this is not very elegant, maybe consider if it's doable to put it into some subfunction?

output/batch/table/formatter.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/olekukonko/tablewriter"
+
+	"github.com/cube2222/octosql"
+	"github.com/cube2222/octosql/execution"
+	"github.com/cube2222/octosql/output/batch"
+)
+
+func TableFormatter(rowLines bool) batch.TableFormatter {
+	return func(w io.Writer, records []*execution.Record, watermark time.Time, errToPrint error) error {
+		var fields []string
```
JasiekChomiak, line 17: This code appears more than once and it's like 12 lines of code - maybe create a function `getAllFields(records []*Record) []string`

output/batch/csv/formatter.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/cube2222/octosql"
+	"github.com/cube2222/octosql/execution"
+	"github.com/cube2222/octosql/output/batch"
+)
+
+func TableFormatter(separator rune) batch.TableFormatter {
+	return func(w io.Writer, records []*execution.Record, watermark time.Time, errToPrint error) error {
+		var fields []string
```
JasiekChomiak, line 17: The function that I mentioned in batch/table/formatter.go should be used here (just a reminder)

output/batch/stdout.go
```diff
@@ -0,0 +1,171 @@
...
+func NewLiveTablePrinter(stateStorage storage.Storage, recordsLister RecordsLister, tableFormatter TableFormatter) *LiveTablePrinter {
+	return &LiveTablePrinter{
+		recordsLister:  recordsLister,
+		stateStorage:   stateStorage,
+		tableFormatter: tableFormatter,
+	}
+}
+
+func (printer *LiveTablePrinter) Run(ctx context.Context) error {
+	for range time.Tick(time.Second / 4) {
```
JasiekChomiak, line 42: This should prolly be some constant/some configurable value

output/batch/stdout.go
```diff
@@ -0,0 +1,171 @@
...
+		tx := printer.stateStorage.BeginTransaction()
+
+		records, err := printer.recordsLister.ListRecords(ctx, tx)
+		if err != nil {
+			return errors.Wrap(err, "couldn't list records")
+		}
+
+		if len(records) > 0 && !records[0].EventTimeField().Empty() {
+			sort.SliceStable(records, func(i, j int) bool {
+				return records[i].EventTime().AsTime().Before(records[j].EventTime().AsTime())
```
JasiekChomiak, line 52: Why do you assume that if the first record has an `EventTimeField` then all records do?

output/batch/stdout.go
```diff
@@ -0,0 +1,171 @@
...
+func NewWholeTablePrinter(stateStorage storage.Storage, recordsLister RecordsLister, tableFormatter TableFormatter) *WholeTablePrinter {
+	return &WholeTablePrinter{
+		recordsLister:  recordsLister,
+		stateStorage:   stateStorage,
+		tableFormatter: tableFormatter,
+	}
+}
+
+func (printer *WholeTablePrinter) Run(ctx context.Context) error {
+	for range time.Tick(time.Second / 4) {
```
JasiekChomiak, line 115: Same as above

output/batch/stdout.go
```diff
@@ -0,0 +1,171 @@
...
+		if !endOfStream && len(errorMessage) == 0 {
+			continue
+		}
+
+		records, err := printer.recordsLister.ListRecords(ctx, tx)
+		if err != nil {
+			return errors.Wrap(err, "couldn't list records")
+		}
+
+		if len(records) > 0 && !records[0].EventTimeField().Empty() {
```
JasiekChomiak, line 137: Same as above

values.go
```diff
@@ -72,6 +72,7 @@ func ZeroString() Value {
 func MakeTime(v time.Time) Value {
 	t, err := ptypes.TimestampProto(v)
 	if err != nil {
+		log.Println(v)
```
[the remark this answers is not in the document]
cube2222: Oh this was actually an accident to leave this here.

sorted_keys.go
```diff
@@ -293,22 +293,30 @@ func MonotonicUnmarshalString(b []byte) (string, error) {
...
 /* Marshal Timestamp */
 func MonotonicMarshalTime(t time.Time) []byte {
-	bytes := MonotonicMarshalInt64(t.UnixNano())
+	bytes := make([]byte, 1+8+1+8)
+
 	bytes[0] = TimestampIdentifier
+	seconds := MonotonicMarshalInt64(t.Unix())
+	copy(bytes[1:9], seconds[1:9]) // Omit integer
+
+	bytes[9] = TimestampIdentifier
```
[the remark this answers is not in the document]
cube2222: The problem is, currently the marshalling/unmarshalling functions are designed so that they accept an array of n+1 bytes and ignore the first one.

The approach of just packing it into 17 bytes works, but really hurts your eyes when reading it. So I opted for this one, which loses a byte, but looks very pleasing.

sorted_keys.go
```diff
@@ -293,22 +293,30 @@ func MonotonicUnmarshalString(b []byte) (string, error) {
 
 /* Marshal Timestamp */
 func MonotonicMarshalTime(t time.Time) []byte {
-	bytes := MonotonicMarshalInt64(t.UnixNano())
+	bytes := make([]byte, 1+8+1+8)
```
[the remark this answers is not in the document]
cube2222: I don't want to remove it.

sorted_keys.go
```diff
@@ -498,7 +506,9 @@ func getConstantMarshalLength(identifier byte) (int, error) {
 	switch identifier {
 	case NullIdentifier, PhantomIdentifier:
 		return NonexistentMarshalLength, nil
-	case IntIdentifier, FloatIdentifier, TimestampIdentifier, DurationIdentifier:
+	case TimestampIdentifier:
+		return 2 * NumberMarshalLength, nil
```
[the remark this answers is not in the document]
cube2222: Don't want to.

sorted_keys.go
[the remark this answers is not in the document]
JasiekChomiak: Oh you are right, forgot about that. Yeah that might not have been the best idea overall. But still for sure change the `1 + 8 + 1 + 8` to `2 * NumberMarshalLength`

output/streaming/stdout.go
```diff
@@ -0,0 +1,72 @@
...
+				log.Println("couldn't listen for changes: ", err)
+			}
+			err = waitableError.Close()
+			if err != nil {
+				log.Println("couldn't close subscription: ", err)
+			}
+			continue
+		} else if err != nil {
+			return errors.Wrap(err, "couldn't get next record")
+		}
```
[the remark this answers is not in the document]
cube2222: I do think this is more readable because with so many cases, I like to see word by word what happens with each case.

app/app.go
cube2222: Fixed, I overlooked this.

cmd/octosql/main.go
cube2222: I used it to write what comes below.

cmd/octosql/main.go
cube2222: See above. Fixed.

cmd/octosql/main.go
cube2222: It won't really help in any way. And this is main, the place where you just construct all dependencies, and come with terms to how ugly it looks. I don't think putting this in a function will help with readability. It's also not used anywhere else.

I named the function type though.

output/batch/stdout.go
cube2222: Because the EventTime is something known for a stream at physical plan time. So we know it holds. If one record in a stream has an event time, then the stream is event timed, then all have it.

I didn't pass the event time in the constructor because this is just minor beautifying of the output, and I may want to rewrite it in the future to be more akin to the stream output (keep the input ordering).

output/batch/stdout.go
cube2222: Same as above

output/batch/stdout.go
JasiekChomiak: OK

cmd/octosql/main.go
JasiekChomiak: OK

sorted_keys.go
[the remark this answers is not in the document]
JasiekChomiak: Right, as above - just change the arithmetics

output/streaming/stdout.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/cube2222/octosql/execution"
+	"github.com/cube2222/octosql/storage"
+)
+
+type StreamPrinter struct {
+	stateStorage storage.Storage
+	recordSink   execution.IntermediateRecordStore
+	printfn      func(record *execution.Record)
```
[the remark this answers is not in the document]
cube2222: maybe printFn? Because Print Function.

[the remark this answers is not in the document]
JasiekChomiak: Still vague, but OK - you can see that it prints records because of the argument, so I guess `printFn` will do the trick

output/streaming/stdout.go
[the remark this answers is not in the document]
JasiekChomiak: I am not a big fan of `continue` but I guess it's not too bad

output/streaming/stdout.go
```diff
@@ -0,0 +1,73 @@
...
+		rec, err := sp.recordSink.Next(ctx, tx)
+		if err == execution.ErrEndOfStream {
+			err := tx.Commit()
+			if err != nil {
+				log.Println("couldn't commit transaction: ", err)
+				continue
+			}
+			break
+		} else if errors.Cause(err) == execution.ErrNewTransactionRequired {
+			if err := tx.Commit(); err != nil {
```
woojtek, line 41: why this Commit is 1liner when `ErrEndOfStream` and `GetErrWaitForChanges` cases use Commit as 2liners?

output/streaming/instant.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return nil
+}
+
+func (o *InstantStreamOutput) Next(ctx context.Context, tx storage.StateTransaction) (*execution.Record, error) {
+	tx = tx.WithPrefix(o.StreamID.AsPrefix())
+	// If no element to get and end of stream then end of stream else wait. Also check error.
+	records := execution.NewOutputQueue(tx.WithPrefix(outputRecordsPrefix))
+	var record execution.Record
+	err := records.Pop(ctx, &record)
```
woojtek, line 51: also maybe newline here, this is kinda too much in 1 place for me

woojtek: or 1 line higher to keep record declaration and Pop together (whichever you prefer)

output/streaming/instant.go
```diff
@@ -0,0 +1,150 @@
...
+		} else if err == storage.ErrNotFound {
+		} else if err != nil {
+			return nil, errors.Wrap(err, "couldn't check end of stream state")
+		}
+
+		return nil, errWaitForChanges
+	} else if err != nil {
+		return nil, errors.Wrap(err, "couldn't get record from records queue")
+	}
+	return &record, nil
```
woojtek, line 79: newline here please

output/streaming/instant.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (o *InstantStreamOutput) UpdateWatermark(ctx context.Context, tx storage.StateTransaction, watermark time.Time) error {
+	return nil
+}
+
+func (o *InstantStreamOutput) GetWatermark(ctx context.Context, tx storage.StateTransaction) (time.Time, error) {
+	panic("not implemented")
```
woojtek, line 87: Will it ever be implemented? Or should it rather be `panic("stream output shouldn't ever be asked about watermark")`

cmd/octosql/main.go
```diff
@@ -117,10 +145,11 @@ With OctoSQL you don't need O(n) client tools or a large data analysis system de
 			storageDirectory = tempDir
 		}
 
+		log.Println(storageDirectory)
```
woojtek, line 148: Delete this log (after you finish testing :P)

cmd/octosql/main.go
```diff
@@ -77,23 +81,47 @@ With OctoSQL you don't need O(n) client tools or a large data analysis system de
...
-			out = csvoutput.NewOutput('\t', os.Stdout)
+		case "stream-json":
+			outputSinkFn = func(stateStorage storage.Storage, streamID *execution.StreamID, eventTimeField octosql.VariableName) (store execution.IntermediateRecordStore, printer output.Printer) {
+				sink := streaming.NewInstantStreamOutput(streamID)
+				output := streaming.NewStreamPrinter(
+					stateStorage,
+					sink,
+					streamingjson.JSONPrinter(),
+				)
+				return sink, output
```
woojtek, line 94: everywhere `out` yet here `output` :P

output/batch/batch.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"io"
+	"time"
+
+	"github.com/cube2222/octosql/execution"
+)
+
+type TableFormatter func(w io.Writer, records []*execution.Record, watermark time.Time, err error) error
+
+func GetAllFields(records []*execution.Record) []string {
```
woojtek, line 12: Wouldn't it be easier and faster and more readable to create a map, then check whether fieldName exist in map if yes then continue if no then put the fieldName into map? Like that you won't need this most inner for loop (just a fairly optimal map lookup)

woojtek: And iterate through map keys to create output

woojtek: I know this is copied code but still... looks quite bad to me :/

woojtek: Oh I see now, fields must be stored in order but this won't be a problem when you store both `fields` and `fieldsMap` so that you just don't need this inner for and it won't be quadratic in time

output/batch/table/formatter.go
```diff
@@ -0,0 +1,43 @@
...
+		fields := batch.GetAllFields(records)
+
+		table := tablewriter.NewWriter(w)
+		table.SetColWidth(64)
+		table.SetRowLine(rowLines)
+		table.SetHeader(fields)
+		table.SetAutoFormatHeaders(false)
+
+		for _, record := range records {
+			var out []string
```
woojtek, line 26: in csv there is `row` and here is `out` - these might have same name imo

output/batch/csv/formatter.go
```diff
@@ -0,0 +1,42 @@
...
+				value := record.Value(octosql.NewVariableName(field))
+				row = append(row, value.Show())
+			}
+			err := out.Write(row)
+			if err != nil {
+				return errors.Wrap(err, "couldn't write row")
+			}
+		}
+
+		out.Flush()
```
woojtek, line 38: no errToPrint printing underneath?

output/batch/stdout.go
```diff
@@ -0,0 +1,170 @@
...
+type LiveTablePrinter struct {
+	recordsLister  RecordsLister
+	stateStorage   storage.Storage
+	tableFormatter TableFormatter
+}
+
+func NewLiveTablePrinter(stateStorage storage.Storage, recordsLister RecordsLister, tableFormatter TableFormatter) *LiveTablePrinter {
+	return &LiveTablePrinter{
+		recordsLister:  recordsLister,
+		stateStorage:   stateStorage,
```
woojtek, line 35: change the order of assignments so it match the arguments :P (im sorry)

output/batch/stdout.go
```diff
@@ -0,0 +1,170 @@
...
+type BatchTablePrinter struct {
+	recordsLister  RecordsLister
+	stateStorage   storage.Storage
+	tableFormatter TableFormatter
+}
+
+func NewWholeTablePrinter(stateStorage storage.Storage, recordsLister RecordsLister, tableFormatter TableFormatter) *BatchTablePrinter {
+	return &BatchTablePrinter{
+		recordsLister:  recordsLister,
+		stateStorage:   stateStorage,
```
woojtek, line 108: same as above

output/batch/stdout.go
```diff
@@ -0,0 +1,170 @@
...
+		tm.Clear()
+		if _, err := tm.Printf(buf.String()); err != nil {
+			return errors.Wrap(err, "couldn't printf output")
+		}
+		tm.Flush()
+
+		tx.Abort()
+
+		if len(errorMessage) > 0 {
+			return errors.New(errorMessage)
```
woojtek, line 89: I think You can just return errorToPrint here

output/batch/stdout.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		if _, err := buf.WriteTo(os.Stdout); err != nil {
+			return errors.Wrap(err, "couldn't print output")
+		}
+
+		tx.Abort()
+
+		if len(errorMessage) > 0 {
+			return errors.New(errorMessage)
```
woojtek, line 164: same as above

output/batch/stdout.go
```diff
@@ -0,0 +1,170 @@
...
+		tm.Flush()
+
+		tx.Abort()
+
+		if len(errorMessage) > 0 {
+			return errors.New(errorMessage)
+		} else if endOfStream {
+			tm.Clear()
+			_, err := buf.WriteTo(os.Stdout)
+			return err
```
woojtek, line 93: no wrap here?

output/batch/csv/formatter.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/cube2222/octosql"
+	"github.com/cube2222/octosql/execution"
+	"github.com/cube2222/octosql/output/batch"
+)
+
+func TableFormatter(separator rune) batch.TableFormatter {
+	return func(w io.Writer, records []*execution.Record, watermark time.Time, errToPrint error) error {
+		if errToPrint != nil {
```
woojtek, line 18: Last 1 thing, in table formatter this is printed in the end of the function and here in the beginning, can You change it so they appear in the same place (and Formatters shapes remain similar)?

cube2222: It's intentional. In practice it fits better because of how they get displayed.

PR #226, opened 2020-05-03
IDs for records in Group By

JasiekChomiak:
This is a small PR that adds adding IDs to records that 
come out of a Group By. I also refactored the record multi set
in test_utils a little bit


execution/group_by.go
```diff
@@ -262,5 +266,28 @@ func (gb *GroupByStream) Trigger(ctx context.Context, tx storage.StateTransactio
...
+		triggeredCount = octosql.MakeInt(0)
+	} else if err != nil {
+		return nil, errors.Wrap(err, "couldn't get count of previously triggered records")
+	}
+
+	triggeredCountValue := triggeredCount.AsInt()
+
+	// Set IDs for records
+	for i := range output {
+		(WithID(NewRecordIDFromStreamIDWithOffset(gb.streamID, triggeredCountValue+i)))(output[i])
```
woojtek, line 283: Aren't there too many parentheses? The one around WithID seems not necessary

PR #232, opened 2020-05-07
Make Triggering Batch Size Based

cube2222:


execution/batch.go
```diff
@@ -66,10 +66,19 @@ func (bsm *BatchSizeManager) ShouldTakeNextRecord() bool {
...
 }
 
+// You can use this to process records in a way other than one by one.
+func (bsm *BatchSizeManager) RecordsLeftToTake() int {
+	return bsm.batchSize - bsm.curBatchSize
+}
+
+// You can use this to process records in a way other than one by one.
+func (bsm *BatchSizeManager) MarkRecordsProcessed(count int) {
+	bsm.curBatchSize += count
```
JasiekChomiak, line 79: Why is this stored in memory?

execution/engine.go
```diff
@@ -107,18 +109,56 @@ func (engine *PullEngine) Run() {
 		default:
 		}
 
-		err := engine.loop(engine.ctx, tx)
+		if engine.batchSizeManager.RecordsLeftToTake() == 0 {
```
JasiekChomiak, line 112: Add a comment here so that someone reading this doesn't have to know what `refreshTransaction` does

execution/process.go
```diff
@@ -156,28 +125,91 @@ func (p *ProcessByKey) Next(ctx context.Context, tx storage.StateTransaction) (*
...
+
+	if len(keys) == 0 {
+		// Send any pending watermark
+		pendingWatermarkState := storage.NewValueState(tx.WithPrefix(pendingWatermarkPrefix))
+		var t timestamp.Timestamp
+		err := pendingWatermarkState.Get(&t)
+		if err == storage.ErrNotFound {
+		} else if err != nil {
+			return 0, errors.Wrap(err, "couldn't get pending watermark state")
+		} else if err == nil {
```
JasiekChomiak, line 161: I think just `else` here should work as well

execution/process.go
```diff
@@ -156,28 +125,91 @@ func (p *ProcessByKey) Next(ctx context.Context, tx storage.StateTransaction) (*
...
+		return 0, errors.Wrap(err, "couldn't poll keys to fire")
 	}
-	err = outputQueue.Push(ctx, &QueueElement{Type: &QueueElement_Watermark{Watermark: timestamp}})
-	if err != nil {
-		return errors.Wrap(err, "couldn't push item to output queue")
+
+	if len(keys) == 0 {
+		// Send any pending watermark
+		pendingWatermarkState := storage.NewValueState(tx.WithPrefix(pendingWatermarkPrefix))
+		var t timestamp.Timestamp
```
woojtek, line 156: could you add a newlne here

execution/process.go
```diff
@@ -156,28 +125,91 @@ func (p *ProcessByKey) Next(ctx context.Context, tx storage.StateTransaction) (*
...
+			}
+
+			if err := pendingWatermarkState.Clear(); err != nil {
+				return 0, errors.Wrap(err, "couldn't clear pending watermark state")
+			}
+		}
+
+		// Send any pending end of stream
+		pendingEndOfStreamState := storage.NewValueState(tx.WithPrefix(pendingEndOfStreamPrefix))
+		var phantom octosql.Value
```
woojtek, line 178: same

execution/process.go
```diff
@@ -156,28 +125,91 @@ func (p *ProcessByKey) Next(ctx context.Context, tx storage.StateTransaction) (*
...
+		err = pendingEndOfStreamState.Get(&phantom)
+		if err == storage.ErrNotFound {
+		} else if err != nil {
+			return 0, errors.Wrap(err, "couldn't get pending end of stream state")
+		} else if err == nil {
```
woojtek, line 183: same as Jasiek said, `else` should be enough here

execution/process.go
```diff
@@ -156,28 +125,91 @@ func (p *ProcessByKey) Next(ctx context.Context, tx storage.StateTransaction) (*
...
+
+			outputQueue := NewOutputQueue(tx.WithPrefix(outputQueuePrefix))
+			if err := outputQueue.Push(ctx, &QueueElement{Type: &QueueElement_EndOfStream{EndOfStream: true}}); err != nil {
```
woojtek, line 186: expand this as with watermark, I really prefer that form :)

woojtek: (Thanks to You btw)

execution/process.go
```diff
@@ -156,28 +125,91 @@ func (p *ProcessByKey) Next(ctx context.Context, tx storage.StateTransaction) (*
...
+		} else if err != nil {
+			return 0, errors.Wrap(err, "couldn't get pending watermark state")
+		} else if err == nil {
+
+			if err := outputQueue.Push(ctx, &QueueElement{
+				Type: &QueueElement_Watermark{
+					Watermark: &t,
+				},
+			}); err != nil {
+				return 0, errors.Wrap(err, "couldn't push record to output queue")
```
woojtek, line 168: maybe couldn't push watermark would be better

execution/batch.go
cube2222: 1. It always has been 2. Because this value doesn't influence the correctness of the final output of the system. So there's no need for it to be transactional or whatever.

execution/process.go
```diff
@@ -19,7 +20,7 @@ type ProcessFunction interface {
 type Trigger interface {
 	RecordReceived(ctx context.Context, tx storage.StateTransaction, key octosql.Value, eventTime time.Time) error
 	UpdateWatermark(ctx context.Context, tx storage.StateTransaction, watermark time.Time) error
-	PollKeysToFire(ctx context.Context, tx storage.StateTransaction) ([]octosql.Value, error)
+	PollKeysToFire(ctx context.Context, tx storage.StateTransaction, batchSize int) ([]octosql.Value, error)
```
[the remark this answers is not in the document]
cube2222: Because you define interfaces where they are used. And you'd get a import loop here otherwise.

PR #235, opened 2020-05-11
Garbage collection

woojtek:


execution/distinct.go
```diff
@@ -13,13 +13,18 @@ type Distinct struct {
 	storage        storage.Storage
 	source         Node
 	eventTimeField octosql.VariableName
+
+	garbageCollectorBoundary int // number of minutes subtracted from watermark to create boundary for garbage collection
+	garbageCollectorCycle    int // number of miliseconds for garbage collector to sleep between records check
```
JasiekChomiak, line 18: Is there a reason why the `garbageCollectorBoundary` is in minutes and `garbageCollectorCycle` in miliseconds? I think the most natural for both would be to store seconds

execution/distinct.go
```diff
@@ -13,13 +13,18 @@ type Distinct struct {
...
 }
 
-func NewDistinct(storage storage.Storage, source Node, eventTimeField octosql.VariableName) *Distinct {
+func NewDistinct(storage storage.Storage, source Node, eventTimeField octosql.VariableName, garbageCollectorBoundary, garbageCollectorCycle int) *Distinct {
```
JasiekChomiak, line 21: Dunno if you agree, but maybe it would be more elegant to create a `type garbageCollectorInfo struct` that contains both these values and then just pass this to the constructor - I think it looks better tbh

execution/group_by.go
```diff
@@ -32,10 +32,13 @@ type GroupBy struct {
 	outEventTimeField octosql.VariableName
 
 	triggerPrototype TriggerPrototype
+
+	garbageCollectorBoundary int // number of minutes subtracted from watermark to create boundary for garbage collection
```
JasiekChomiak, line 36: Following up on the above - if you added such a `struct` then these comments explaining which field means what, would only need to appear once in the `struct` definition

execution/group_by.go
```diff
@@ -280,7 +302,11 @@ func (gb *GroupByStream) Trigger(ctx context.Context, tx storage.StateTransactio
 
 	// Set IDs for records
 	for i := range output {
-		WithID(NewRecordIDFromStreamIDWithOffset(gb.streamID, triggeredCountValue+i))(output[i])
+		keyHash, err := key.Hash()
```
JasiekChomiak, line 305: Could you explain what was wrong with the previous ID setting?

execution/stream_join.go
```diff
@@ -38,18 +38,23 @@ type StreamJoin struct {
 	eventTimeField          octosql.VariableName
 	joinType                JoinType
 	triggerPrototype        TriggerPrototype
+
+	garbageCollectorBoundary int // number of minutes subtracted from watermark to create boundary for garbage collection
```
JasiekChomiak, line 42: Same as above

execution/process.go
```diff
@@ -54,16 +59,43 @@ func (p *ProcessByKey) AddRecord(ctx context.Context, tx storage.StateTransactio
...
 	if len(p.eventTimeField) > 0 {
 		eventTime = record.EventTime().AsTime()
+
+		// Adding record event time to keyTuple so that during Trigger prefixing process function is possible
+		keyTuple = octosql.MakeTuple(append([]octosql.Value{octosql.MakeTime(eventTime)}, key...))
+
+		// Adding keyTuple to event time storage
+		eventTimeMap := storage.NewMap(tx.WithPrefix(eventTimesSeenPrefix))
+
+		octoEventTime := octosql.MakeTime(eventTime)
```
JasiekChomiak, line 74: This is not entirely clear. You could add a comment explaining it a bit more like "for every event time, we store keys that appeared in that event time" or smth like that

execution/process.go
```diff
@@ -54,16 +59,43 @@ func (p *ProcessByKey) AddRecord(ctx context.Context, tx storage.StateTransactio
...
+		var recordsTuple octosql.Value
```
JasiekChomiak, line 75: Shouldn't this be called `keysTuple` since it stores keys?

execution/process.go
```diff
@@ -54,16 +59,43 @@ func (p *ProcessByKey) AddRecord(ctx context.Context, tx storage.StateTransactio
...
+		eventTimePrefixedTx = tx.WithPrefix(eventTimeBytes)
+	}
+
+	// Here garbage collector will just drop event time prefixes
+	err = p.processFunction.AddRecord(ctx, eventTimePrefixedTx, inputIndex, keyTuple, record)
+	if err != nil {
+		return errors.Wrap(err, "couldn't add record to process function")
 	}
 
+	// Here garbage collector will call KeysFired so that trigger itself can remove records from storage
```
JasiekChomiak, line 98: The comment says it calls `KeysFired` but it calls `RecordReceived`?

execution/process.go
```diff
@@ -72,6 +104,69 @@ func (p *ProcessByKey) AddRecord(ctx context.Context, tx storage.StateTransactio
...
+
+		// Check current output watermark value
+		watermark, err := p.GetWatermark(ctx, tx)
+		if err != nil {
+			return errors.Wrap(err, "couldn't get output watermark value")
+		}
+
+		// We don't want to clear whole storage when sending MaxWatermark - on EndOfStream and WatermarkTrigger
+		if watermark != MaxWatermark {
+			// Collect every event time earlier than watermark - 10min
```
JasiekChomiak, line 125: 10min is misleading - it's `watermark - boundary`

execution/group_by.go
woojtek: record id's were duplicated, because every storage in processFunction is prefixed with event time, so `triggeredCount` storage also is which means that for given records with event times
`1 1 1 2 2 3 3`
instead of record ids
`0 1 2 3 4 5 6`
you receive
`0 1 2 0 1 0 1`
that's why there is key hass added to record id to prevent duplications

execution/process.go
```diff
@@ -54,16 +59,45 @@ func (p *ProcessByKey) AddRecord(ctx context.Context, tx storage.StateTransactio
...
+		octoEventTime := octosql.MakeTime(eventTime)
+		var keysTuple octosql.Value
+		err := eventTimeMap.Get(&octoEventTime, &keysTuple)
+		if err == storage.ErrNotFound {
+			keysTuple = octosql.ZeroTuple()
+		} else if err != nil {
+			return errors.Wrap(err, "couldn't get records event time tuple from storage")
+		}
+
+		keysTuple = octosql.MakeTuple(append(keysTuple.AsSlice(), keyTuple.AsSlice()...))
```
cube2222, line 85: Firstly, I don't udnerstand why you add the three dots here. This would separate the key tuple into each value it contains and append them to the keysTuple slice. Wouldn't you want each element to be one full key tuple?

cube2222, line 85: Secondly, this will completely decimate performance. The slice will grow for each key encountered, and each time you will load and set all keys. Please create a Set (we have those) prefixed by the event time.

execution/process.go
```diff
@@ -261,5 +303,87 @@ func (p *ProcessByKey) ReadyForMore(ctx context.Context, tx storage.StateTransac
 }
 
 func (p *ProcessByKey) Close(ctx context.Context, storage storage.Storage) error {
+	p.garbageCollectorCtxCancel()
+	err := <-p.garbageCollectorCloseErrChan
+	if err == context.Canceled || err == context.DeadlineExceeded {
+	} else if err != nil {
+		return errors.Wrap(err, "couldn't stop garbage collector")
+	}
```
cube2222, line 311: Shouldn't this close function also drop all local data? (DropPrefix?)

execution/process.go
```diff
@@ -261,5 +303,87 @@ func (p *ProcessByKey) ReadyForMore(ctx context.Context, tx storage.StateTransac
...
+
 	return nil
 }
+
+type GarbageCollectorInfo struct {
+	garbageCollectorBoundary int // number of seconds subtracted from watermark to create boundary for garbage collection
```
cube2222, line 317: If you're operating on duration values, please use durations, not int of seconds.

execution/process.go
```diff
@@ -261,5 +303,87 @@ func (p *ProcessByKey) ReadyForMore(ctx context.Context, tx storage.StateTransac
...
+	garbageCollectorCycle    int // number of seconds for garbage collector to sleep between records check
```
cube2222, line 318: same

execution/process.go
woojtek: It should but only on storages used in methods `GetWatermark` and `Next` as rest of methods are called with engine prefixed tx. Will fix this

woojtek: Actually, now `Next` must be cleared separately because `StreamPrinter` calls `Next` without prefixing it...

woojtek: but storage used in GarbageCollector doesn't have to be cleared as it's creating transaction from prefixed (with pull engine streamID) storage, so pull engine will clear it

execution/process.go
woojtek: Yes, You're right

execution/process.go
```diff
@@ -344,46 +357,74 @@ func (p *ProcessByKey) RunGarbageCollector(ctx context.Context, prefixedStorage
...
+					eventTimeBytes := append(append([]byte("$"), octoEventTime.MonotonicMarshal()...), '$')
+
+					octoEventTimeSlice = append(octoEventTimeSlice, octoEventTime)
+					eventTimesBytesSlice = append(eventTimesBytesSlice, eventTimeBytes)
 
-					eventTimeBytes := append(append([]byte("$"), key.MonotonicMarshal()...), '$')
+					// Iterating through all keyTuples with that event time
+					keyTuplesSet := storage.NewMultiSet(tx.WithPrefix(eventTimesPrefixedSetsPrefix).WithPrefix(eventTimeBytes))
+
+					allKeysTuple := make([]octosql.Value, 0)
```
JasiekChomiak, line 382: MultiSet has a `ReadAll` function. Use it here please

execution/process.go
```diff
@@ -310,22 +319,26 @@ func (p *ProcessByKey) Close(ctx context.Context, storage storage.Storage) error
...
+func NewGarbageCollectorInfo(garbageCollectorBoundary, garbageCollectorCycle time.Duration) *GarbageCollectorInfo {
 	return &GarbageCollectorInfo{
 		garbageCollectorBoundary: garbageCollectorBoundary,
 		garbageCollectorCycle:    garbageCollectorCycle,
 	}
 }
 
-func (p *ProcessByKey) RunGarbageCollector(ctx context.Context, prefixedStorage storage.Storage, gbBoundary, gbCycle int) error {
+func (p *ProcessByKey) RunGarbageCollector(ctx context.Context, prefixedStorage storage.Storage, gbBoundary, gbCycle time.Duration) error {
 	for {
```
JasiekChomiak, line 342: Instead of the `sleep` at the very end maybe it's more elegant to do `for range time.Tick(gbCycle * time.Second)`

woojtek: absolutely right, completely forgot about this since i've never used tick before :P

execution/process.go
woojtek: oooh yes, that's a new one, lovely

PR #240, opened 2020-05-28
Order by streaming

Risers:


execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+
+func (k *OrderByKey) MonotonicUnmarshal(data []byte) error {
+	k.key = make([]byte, len(data))
+	copy(k.key, data)
+	return nil
+}
+
+
 type OrderBy struct {
+	storage storage.Storage
```
woojtek, line 34: please use reformat shortcut to indent struct fields properly

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
 	}
 
-	defer func() {
-		if err := recover(); err != nil {
-			stream = nil
-			outErr = errors.Wrap(err.(error), "couldn't sort records")
+	keyPrefix := append(append([]byte("$"), key.MonotonicMarshal()...), '$')
+	txByKey := tx.WithPrefix(keyPrefix)
+
+	var recordValueMap *storage.Map
```
JasiekChomiak, line 113: Why not simply `recordValueMap := storage.NewMap(...)`?

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+	var recordCountMap *storage.Map
```
JasiekChomiak, line 114: Same as above

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
 
-	for {
-		rec, err := sourceStream.Next(ctx)
-		if err == ErrEndOfStream {
-			break
-		} else if err != nil {
-			return nil, errors.Wrap(err, "couldn't get all records")
-		}
+var recordValuePrefix = []byte("$record_value$")
+//todo record count prefix is from orderby
```
woojtek, line 103: is this todo old or still not resolved?

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+
+	bytePref := append(append(recordPrefix, []byte(NewRecordFromRecord(record, WithNoUndo()).String())...), '$')
+	pref := OrderByKey{key: bytePref}
+
+	count  := octosql.MakePhantom()
+	err := recordCountMap.Get(&pref, &count)
+
+	if !record.IsUndo() {
+		if err == storage.ErrNotFound {
+
```
JasiekChomiak, line 150: Unnecessary newline

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+			err = recordValueMap.Set(&pref, record)
 			if err != nil {
-				panic(errors.Wrap(err, "couldn't merge variables"))
+				return err
```
JasiekChomiak, line 153: Wrap this error please

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
 			}
 
-			x, err := expr.ExpressionValue(ctx, iVars)
+			count = octosql.MakeInt(1)
+			err = recordCountMap.Set(&pref, &count)
 			if err != nil {
-				panic(errors.Wrapf(err, "couldn't get order by expression with index %v value", num))
+				return err
```
JasiekChomiak, line 159: Same as above

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
 			}
-			y, err := expr.ExpressionValue(ctx, jVars)
+		} else if err != nil {
+			return err
```
JasiekChomiak, line 162: Same as above

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+		} else {
+			// Keep track of record vs retraction count
+			count = octosql.MakeInt(count.AsInt() + 1)
+			err = recordCountMap.Set(&pref, &count)
 			if err != nil {
-				panic(errors.Wrapf(err, "couldn't get order by expression with index %v value", num))
+				return err
```
JasiekChomiak, line 168: Same as above

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
-				panic(errors.Errorf("value %v of type %v is not comparable", x, reflect.TypeOf(x).String()))
-			}
-			if !isSorteable(y) {
-				panic(errors.Errorf("value %v of type %v is not comparable", y, reflect.TypeOf(y).String()))
+		}
+	} else {
+		if err == storage.ErrNotFound {
+			return errors.Wrap(err, "couldn't retract record")
+		} else if err != nil {
+			return err
```
JasiekChomiak, line 175: Same as above

execution/test_utils.go
```diff
@@ -409,6 +409,35 @@ func AreStreamsEqualNoOrdering(ctx context.Context, stateStorage storage.Storage
...
+	log.Println("read first stream")
+
+	log.Println("second stream")
+	secondRecords, err := ReadAll(ctx, stateStorage, second)
+	if err != nil {
+		return errors.Wrap(err, "couldn't read second streams records")
+	}
+	log.Println("read second stream")
+
+	if len(firstRecords) == len(secondRecords) {
```
woojtek, line 428: why `==` and not `!=` ?

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+			count = octosql.MakeInt(count.AsInt() - 1)
+			if count.AsInt() < 0 {
+				return errors.New( "couldn't retract record")
 			}
 
-			cmp, err := octosql.Compare(x, y)
+			err = recordCountMap.Set(&pref, &count)
 			if err != nil {
-				panic(errors.Errorf("failed to compare values %v and %v", x, y))
+				return err
```
JasiekChomiak, line 185: Same as above

execution/test_utils.go
woojtek: if they are of equal length then error should be `streams don't have different ammount of records` or `streams have equal amount of records`

woojtek: amount* not ammount

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+	}
+	variables := record.AsVariables()
+
+
+	recordPrefix := []byte("$")
+
+	for i := range ob.expressions {
+		expressionValue, err := ob.expressions[i].ExpressionValue(ctx, variables)
+		if err != nil {
+			return errors.Wrap(err, "couldn't evaluate expression")
```
JasiekChomiak, line 132: You could use `Wrapf` and include the index of the expression like "couldn't evalute expression with index %d", i

sorted_keys.go
```diff
@@ -42,6 +42,15 @@ func (v *Value) MonotonicMarshal() []byte {
 	return monotonicMarshal(v)
 }
 
+func (v *Value) ReversedMonotonicMarshal() []byte {
```
woojtek, line 45: wouldn't `XORedMonotonicMarshal` be a more appropriate name?

physical/order_by.go
```diff
@@ -58,6 +59,10 @@ func (node *OrderBy) Materialize(ctx context.Context, matCtx *MaterializationCon
 		}
 	}
 
+
+	var triggerPrototype execution.TriggerPrototype
+	triggerPrototype = execution.NewWatermarkTrigger()
```
woojtek, line 64: replace these 2 lines with `triggerPrototype := execution.NewWatermarkTrigger()`

physical/order_by.go
```diff
@@ -68,7 +73,23 @@ func (node *OrderBy) Materialize(ctx context.Context, matCtx *MaterializationCon
 		return nil, errors.Wrap(err, "couldn't get execution node from order by source")
 	}
 
-	return execution.NewOrderBy(exprs, directions, sourceNode), nil
+	eventTimeField := octosql.NewVariableName("")
+
+	if node.orderByEventTime(node.Source.Metadata()) {
+		eventTimeField = node.Expressions[0].(*Variable).ExpressionName()
```
woojtek, line 79: This could be done better, as You need to parse `node.Expression[0]` twice.
Either delete that function and perform everything in this function or return `sourceMetadata.EventTimeField()` instead of boolean

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
-		if err := recover(); err != nil {
-			stream = nil
-			outErr = errors.Wrap(err.(error), "couldn't sort records")
+	keyPrefix := append(append([]byte("$"), key.MonotonicMarshal()...), '$')
+	txByKey := tx.WithPrefix(keyPrefix)
+
+	var recordValueMap *storage.Map
+	var recordCountMap *storage.Map
+
+
```
woojtek, line 116: too many empty lines

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+	recordValueMap = storage.NewMap(txByKey.WithPrefix(recordValuePrefix))
+	recordCountMap = storage.NewMap(txByKey.WithPrefix(recordCountPrefix))
+
+	mapping := make(map[string]octosql.Value, len(record.Fields()))
+	for _, field := range record.Fields() {
+		mapping[field.Name.String()] = record.Value(field.Name)
+	}
+	variables := record.AsVariables()
+
+
```
woojtek, line 126: too many empty lines

execution/order_by.go
woojtek, line 114: replace with

```
recordValueMap := storage.NewMap(txByKey.WithPrefix(recordValuePrefix))
recordCountMap := storage.NewMap(txByKey.WithPrefix(recordCountPrefix))
```

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+		} else {
+			recordPrefix = append(recordPrefix, expressionValue.ReversedMonotonicMarshal()...)
+		}
+		recordPrefix = append(recordPrefix, '$')
+	}
+
+	bytePref := append(append(recordPrefix, []byte(NewRecordFromRecord(record, WithNoUndo()).String())...), '$')
+	pref := OrderByKey{key: bytePref}
+
+	count  := octosql.MakePhantom()
```
JasiekChomiak, line 145: Please use a `ValueState` to keep track of record count - you can look at how this is implemented in a Group By or Stream Join

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
-			outErr = errors.Wrap(err.(error), "couldn't sort records")
+	keyPrefix := append(append([]byte("$"), key.MonotonicMarshal()...), '$')
+	txByKey := tx.WithPrefix(keyPrefix)
+
+	var recordValueMap *storage.Map
+	var recordCountMap *storage.Map
+
+
+	recordValueMap = storage.NewMap(txByKey.WithPrefix(recordValuePrefix))
+	recordCountMap = storage.NewMap(txByKey.WithPrefix(recordCountPrefix))
```
woojtek, line 118: where is this prefix defined?

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+func (ob *OrderByStream) Trigger(ctx context.Context, tx storage.StateTransaction, key octosql.Value) ([]*Record, error) {
+	output := make([]*Record, 0)
 
-			if directions[num] == Ascending {
-				answer = !answer
-			}
+	keyPrefix := append(append([]byte("$"), key.MonotonicMarshal()...), '$')
+	txByKey := tx.WithPrefix(keyPrefix)
+
+	var recordValueMap *storage.Map
```
JasiekChomiak, line 199: Same as in AddRecord, why not just use `:=`

sorted_keys.go
woojtek: ok I get it now, it's OK but please add some comment what is it used for (here or in the order by file)

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
-			jVars, err := variables.MergeWith(jRec.AsVariables())
+		if ob.directions[i] == Ascending {
+			recordPrefix = append(recordPrefix, expressionValue.MonotonicMarshal()...)
+		} else {
+			recordPrefix = append(recordPrefix, expressionValue.ReversedMonotonicMarshal()...)
+		}
+		recordPrefix = append(recordPrefix, '$')
+	}
+
+	bytePref := append(append(recordPrefix, []byte(NewRecordFromRecord(record, WithNoUndo()).String())...), '$')
```
woojtek, line 142: please add a comment explaining why you need to prefix whole record, it's not clear for me

execution/order_by.go
woojtek, line 162: wrap errors

execution/order_by.go
woojtek, line 168: same, wrap errors (ALL errors, maybe I missed some)

execution/order_by.go
woojtek, line 153: wrap errors

execution/order_by.go
woojtek, line 159: wrap errors

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
-
-			if !isSorteable(x) {
-				panic(errors.Errorf("value %v of type %v is not comparable", x, reflect.TypeOf(x).String()))
-			}
-			if !isSorteable(y) {
-				panic(errors.Errorf("value %v of type %v is not comparable", y, reflect.TypeOf(y).String()))
+		}
+	} else {
+		if err == storage.ErrNotFound {
+			return errors.Wrap(err, "couldn't retract record")
```
woojtek, line 173: This is a tricky case, but You should handle early retractions here, which means save count -1 instead of returning error

execution/order_by.go
woojtek, line 175: wrap

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+		} else if err != nil {
+			return err
+		} else {
+			// Keep track of record vs retraction count
+			count = octosql.MakeInt(count.AsInt() - 1)
+			if count.AsInt() < 0 {
+				return errors.New( "couldn't retract record")
```
woojtek, line 180: same as above, early retractions, dont return error, just save negative count value

execution/order_by.go
woojtek, line 185: wrap

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
-			}
+	keyPrefix := append(append([]byte("$"), key.MonotonicMarshal()...), '$')
+	txByKey := tx.WithPrefix(keyPrefix)
+
+	var recordValueMap *storage.Map
+	var recordCountMap *storage.Map
 
-			return answer
+	recordValueMap = storage.NewMap(txByKey.WithPrefix(recordValuePrefix))
+	recordCountMap = storage.NewMap(txByKey.WithPrefix(recordCountPrefix))
```
woojtek, line 203: same as above, use `:=` construction

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+
+	var err error
+	pref := OrderByKey{key: []byte("")}
```
woojtek, line 206: I would write this like that
```
pref := OrderByKey{key: []byte("")}
<empty line>
var count octosql.Value
var value Record
var err error
```

just separate nil values from created prefix

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+	iter := recordValueMap.GetIterator()
+	for err = iter.Next(&pref, &value); err == nil; err = iter.Next(&pref, &value) {
+		err := recordCountMap.Get(&pref, &count)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't get record count")
+		}
+
+		for i := 0; i < count.AsInt(); i++ {
+			newValue := value
+			output = append(output, &newValue)
```
woojtek, line 218: `newValue` is not needed here, just use `&value`

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
 		}
+	}
 
-		return false
-	})
+	if err != storage.ErrEndOfIterator {
+		return nil, errors.Wrap(err, "couldn't iterate over existing records")
+	}
+
```
woojtek, line 225: no empty line here :)

execution/order_by.go
```diff
@@ -18,132 +15,217 @@ const (
...
+type OrderByKey struct {
+	key []byte
+}
+
+func (k *OrderByKey) MonotonicMarshal() []byte {
+	return k.key
+}
+
+func (k *OrderByKey) MonotonicUnmarshal(data []byte) error {
+	k.key = make([]byte, len(data))
```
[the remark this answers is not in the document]
Risers: I can't because then it copies reference and not value

execution/order_by.go
Risers: it is needed because we need new reference for each value in output

woojtek: ok

Risers:
> I'm okay with the changes now, but 1 last thing I didn't think about previously, we are holding recordcount and recordvalue separately, couldn't we just keep a single map `recordValue -> recordCount`?

We probably could do it but then but then we would have to extract value from serialized map key, which is more than just record value. It may decrease code readability.

PR #254, opened 2020-10-03
WIP: Rust First/Last aggregates

woojtek:
"thread 'main' panicked at 'bug: key doesn't match schema', src\physical\group_by.rs:381:36"
while running "SELECT c.livesleft, FIRST(c.id) FROM cats.csv c GROUP BY c.livesleft"

### DIFF
--- a/src/physical/aggregate.rs
+++ b/src/physical/aggregate.rs
@@ -20,6 +20,7 @@ use anyhow::Result;
 use arrow::datatypes::DataType;
 
 use crate::physical::physical::ScalarValue;
+use arrow::ipc::Utf8;
 
 pub trait Aggregate: Send + Sync {
     fn create_accumulator(&self, input_type: &DataType) -> Box<dyn Accumulator>;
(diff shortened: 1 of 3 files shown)


src/physical/aggregate.rs
```diff
@@ -343,3 +343,94 @@ impl_max_accumulator!(u8, UInt8);
...
+    fn create_accumulator(&self, input_type: &DataType) -> Box<dyn Accumulator> {
+        match input_type {
+            DataType::Int8 => Box::new(FirstAccumulator::<i8> { values_order: VecDeque::new(), values_counts: BTreeMap::new() }),
+            DataType::Int16 => Box::new(FirstAccumulator::<i16> { values_order: VecDeque::new(), values_counts: BTreeMap::new() }),
+            DataType::Int32 => Box::new(FirstAccumulator::<i32> { values_order: VecDeque::new(), values_counts: BTreeMap::new() }),
+            DataType::Int64 => Box::new(FirstAccumulator::<i64> { values_order: VecDeque::new(), values_counts: BTreeMap::new() }),
+            DataType::UInt8 => Box::new(FirstAccumulator::<u8> { values_order: VecDeque::new(), values_counts: BTreeMap::new() }),
+            DataType::UInt16 => Box::new(FirstAccumulator::<u16> { values_order: VecDeque::new(), values_counts: BTreeMap::new() }),
+            DataType::UInt32 => Box::new(FirstAccumulator::<u32> { values_order: VecDeque::new(), values_counts: BTreeMap::new() }),
+            DataType::UInt64 => Box::new(FirstAccumulator::<u64> { values_order: VecDeque::new(), values_counts: BTreeMap::new() }),
```
cube2222, line 359: Please also make it work for DataType::Utf8, DataType::Timestamp and DataType:Struct.

They're used in group_by.rs so you can look there for an example. (Utf8 requires a seperate macro unfortunately, as you have to do .as_str() instead of .clone() on the value you append to the builder)

cube2222: Oh, and Boolean, as well as Null.

cube2222: You can ignore Struct for now, because implementing that as a map key could be hard.

cube2222: You can omit structs for now TBH, because creating a builder for them could be hard.

woojtek: I don't understand thata macro builder stuff with UTF8 and also the generic types for null and utf8, so I left them commented maybe you can do something with it :<

PR #259, opened 2021-02-16
Get JSON from API

brenzel:
This change allows to get json from api if the path starts with http

### DIFF
--- a/datasources/json/datasource.go
+++ b/datasources/json/datasource.go
@@ -8,6 +8,8 @@ import (
 	"os"
 	"sort"
 	"time"
+	"strings"
+	"net/http"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


datasources/json/datasource.go
```diff
@@ -154,16 +156,28 @@ func (rs *RecordStream) RunWorker(ctx context.Context) error {
...
-		file, err := os.Open(rs.filePath)
-		if err != nil {
-			return errors.Wrap(err, "couldn't open file")
+			r, err := httpClient.Get(rs.filePath)
+			if err != nil {
+				return errors.Wrap(err, "couldn't open api")
+			}
+			defer r.Body.Close()
+			rs.decoder = json.NewDecoder(r.Body)
+		} else {
```
cube2222, line 171: You're not setting the rs.file, which is getting closed in the .Close method, this will segfault.

You should still set it, and probably change the rs.file to be an io.Closer.

cube2222 (changes requested):
Hey, please create an issue before creating a Pull Request to discuss planned changes first.

In general, the idea is ok, though there are some parts to fix.

I'm not sure why the last CI/CD step is failing, the tests pass however, so no worries about that.

Overall though, I'm not sure if you want to keep working on this (but feel free to do the fixes as I commented if you want to, I will accept it) as I'm in midst of a full rewrite of OctoSQL, so these changes will be discarded in a matter of weeks, when the rewrite will be merged (all current files in the codebase will be deleted).

PR #308, opened 2022-12-08
Switch $HOME/.octosql to XDG Spec

Tony-Sol:
Change default configs/plugins/repos/logs location according to XDG Specification.

App will look for:
- the config file in `$XDG_CONFIG_HOME/octosql/octosql.yml` with fallback to current `$HOME/.octosql/octosql.yml`
- the logs file in `$XDG_CACHE_HOME/octosql/logs.txt` with fallback to current `$HOME/.octosql/logs.txt`
- the plugins tmp dir in `$XDG_CACHE_HOME/octosql/tmp/plugins/` with fallback to current `$HOME/.octosql/tmp/plugins/`
- the extension_handlers file in `$XDG_CONFIG_HOME/octosql/file_extension_handlers.json` with fallback to current `$HOME/.octosql/file_extension_handlers.json`
- the plugins dir in `$XDG_DATA_HOME/octosql/plugins` with fallback to current `$HOME/.octosql/plugins`
- the repositories dir in `$XDG_DATA_HOME/octosql/repositories` with fallback to current `$HOME/.octosql/repositories`

PR #315, opened 2023-02-25
Ensure octosql home directory exists before return

Tony-Sol:
Fixes for https://github.com/cube2222/octosql/pull/308


cube2222 (approved):

cube2222:
Thanks! This looks much better! @Tony-Sol

PR #316, opened 2023-02-26
Prevent creating ~/.octosql dir if XDG dirs exists

Tony-Sol:
Sorry for another fix, but this PR will prevent creating ~/.octosql dir if XDG dirs exist by lateinit of octosqlHomeDir


cube2222 (approved):

cube2222:
Looks good!

PR #322, opened 2023-04-30
update badge url as per https://github.com/badges/shields/issues/8671

phanirithvij:


cube2222 (approved):
Thanks a lot @phanirithvij!

cube2222 (approved):
Thanks a lot @phanirithvij!

PR #331, opened 2024-04-23
Add int64 support

### LINKED ISSUE #330: Add Int64 type
Hey there,

I wanted to spend some time to analyse byte sizes in etcd databases with an [octosql plugin I wrote](https://github.com/tjungblu/octosql-plugin-etcdsnapshot). 

Unfortunately, there's only 32 bit Int, which easily overflows. Would you be interested to have a 64 bit integer as a type for octosql?

tjungblu:
fixes #330

@cube2222 I hope you're still alive. I'm seriously bitten by the lack of int64 for byte sizes and other large aggregations, so it would be incredible if we could add this one way or the other.

### DIFF
--- a/aggregates/average.go
+++ b/aggregates/average.go
@@ -15,6 +15,11 @@ var AverageOverloads = []physical.AggregateDescriptor{
 		Prototype:    NewAverageIntPrototype(),
 	},
 	{
+		ArgumentType: octosql.Int64,
+		OutputType:   octosql.Int64,
+		Prototype:    NewAverageInt64Prototype(),
+	},
+	{
 		ArgumentType: octosql.Float,
 		OutputType:   octosql.Float,
 		Prototype:    NewAverageFloatPrototype(),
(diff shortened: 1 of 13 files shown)


cube2222:
Hey @tjungblu, alive and well! Just didn't have time for OctoSQL for a while now.

Since you're hitting this issue, are you on a 32-bit system? My assumption with OctoSQL was mostly building for 64-bit systems, where int would be 64-bit anyways.

I have fairly intentionally avoided different bit type variants in the octosql type system to avoid the complexity. What would you think, instead, about changing all the places that are problematic where simple `int` is used, to just use int64? Basically, what you did in your PR, but instead of creating new structures, modifying the existing ones to just always use int64. Would that work?

tjungblu:
You're correct, int on 64 bits on a 64 bit system:

```
$ octosql "SELECT 2147483647+2147483647, 9223372036854775807+9223372036854775807"
+------------+-------+
|   col_0    | col_1 |
+------------+-------+
| 4294967294 |    -2 |
+------------+-------+
```

On a raspberry pi armv7l, which is 32 bit it's unfortunately not:

```
$ octosql "SELECT 2147483647+2147483647, 9223372036854775807+9223372036854775807"
+-------+-------+
| col_0 | col_1 |
+-------+-------+
|    -2 |    -2 |
+-------+-------+
```

I realize there's no official 32 bit build of octosql, but it's somewhat inconvenient you have to be aware of your processor to get the correct result :roll_eyes: 

> but instead of creating new structures, modifying the existing ones to just always use int64. Would that work?

let me try! I'll send you another follow-up PR, we can decide what to make of it.

Thanks @cube2222 :)

